Issue #6, opened 2021-02-05
Create a Style Guide


JasonMorgan:
Build, publish, and maintain a style guide so that new folks can more easily contribute.

JasonMorgan:
A couple items I'd like to see, the big one being a definitive template for entries and an explanation for every property. An example: I put in a category property for each entry but as of right now the only value I can think of is concept. Either we should have more categories or remove that property.

IdealUsrname:
I've completed a basic outline for a style guide. Just waiting for the acceptance of #31, So I don't have to change the format again :)

Issue #5, opened 2021-02-05
Reformat the glossary.


JasonMorgan:
Turns out a single document works well in google docs but it doesn't fit super well for a git repo. 

I propose we break the glossary down into a series of individual entries. I further propose we move from markdown to yaml for the entries so they're easier to stitch together into a final product. I've got a PR incoming that will demonstrate my intent.

JasonMorgan:
@CathPag FYI

JasonMorgan:
I set up a template for the individual entries.

```yaml
title: Definition Template
category: concept 
status: Feedback Appreciated # Valid values are: In progress, Completed, Feedback Appreciated, Not Started
definition: |
  A Quick summary of the concept, what it is.

  A few lines of the problem it's addressing.

  A few lines of how the thing solves the problem.
```

I'm going to take a look at a couple other places that have technical definitions. I think it might be handy to have a couple other properties, Like related terms, or see also.

Thanks,
Jason

PR #7, opened 2021-02-05
Removed definitions from the glossary doc.

JasonMorgan:
Trying to ensure it's simple to find the particular definition you care about by pulling definitions into their own individual files.


JasonMorgan:
Intended to address issue #5.

IdealUsrname (commented):
I see no issues with these changes. The individual files allow for fewer conflicts moving forward.

If I may suggest, any further format changes can be added to the "_style guide"_ first. Once that is defined, we can reevaluate and apply the style changes.

Thoughts? @JasonMorgan @CathPag

JasonMorgan:
100% agree, raised issue #6 for that.

PR #9, opened 2021-02-07
Adding "contributing" doc

CathPag:
Based on the State-of-the-Edge doc as suggested by Chris. @JasonMorgan @idealusrname, let me know what you think.


CathPag:
I missed something. Can we add this?

Format 
Please write a glossary definition with the following format: (1) what it is, (2) problem it addresses, and (3) how it helps. Each section shouldn't exceed 100 words (a total of 300 words max per term).

caniszczyk:
I'm going to merge this is now, we can make some small changes to it later like pointing to where the list of maintainers are etc

Issue #16, opened 2021-02-09
Continuous delivery (CD) feedback not yet woven in


CathPag:
We got some great feedback in the Google doc that was not woven into [this CD entry](https://github.com/cncf/glossary/blob/main/definitions/continuous_delivery.yaml). Copying and pasting in below. Please update entry accordingly. 

**Daniel Jones:** I think somewhere in here it'd be great to call out specifically that the path to production (or an acceptance environment) must be fully automated, and also that there process changes involved to - for example, trunk-based development rather than feature branching.

**Michael Foster**: Agreed. Also, thoughts on referencing CI? CI gets you part way there but CI/CD gets you a fully automated workflow. Or we could have a CI/CD definition?

**Daniel Jones**: I'd be open to referencing CI in the CD definition, and vice versa, in order to explain them via contrast. Any objections?

TheJaySmith:
I can take it

IdealUsrname:
With #67, I'm assuming this issue can be closed. Feel free to reopen if the CD conversation sparks back up again.

PR #29, opened 2021-02-11
Continuous delivery (CD) feedback not yet woven in #16

TheJaySmith:
I went ahead and updated the text


JasonMorgan (approved):
Looks awesome! We have to be sure we de conflict this with PR #31 cause I just changed all the entries to markdown and that'll override these changes.

PR #33, opened 2021-02-14
Update _TEMPLATE.md

CathPag:
Deleted "Status: Feedback Appreciated" below the metadata as that info is already part of the metadata. Unless there is a reason I'm not aware of, I'll go ahead and delete it in all other files and commit directly to the main branch once this is merged.

### DIFF
--- a/definitions/_TEMPLATE.md
+++ b/definitions/_TEMPLATE.md
@@ -5,8 +5,6 @@ category: concept
 ---
 ## Definition Template
 
-**Status:** Feedback Appreciated
-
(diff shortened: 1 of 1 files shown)


JasonMorgan:
I don't mind removing it but then we'd need to remove it on all of them. Thoughts?

CathPag:
That's what I said above :) Unless there is a reason, I'll go ahead and remove them all.

IdealUsrname:
Sounds good. I'll update the style guide and create a PR later today.

PR #38, opened 2021-02-17
First pass for the style guide

IdealUsrname:
Closes #6 

I added some extra categories. I would love feedback on formatting.

PR #39, opened 2021-02-17
mfoster/removed status

IdealUsrname:
Hi!

- I removed the status section from all of the definitions.  

- Should not conflict with #38 (warrants a squash and merge)

- Is the continuation to PR #33.


CathPag (approved):
Perfect!

Issue #14, opened 2021-02-09
Need data for Auto Scaling


JasonMorgan:
This [entry](https://github.com/cncf/glossary/blob/main/definitions/auto_scaling.yaml) is blank. We need someone to add a definition.

PR #40, opened 2021-02-18
added autoscaling definition. removed access yaml

IdealUsrname:
Solves #14.

Thoughts on Autoscaling vs. Auto Scaling?

Personally leaning towards Autoscaling due to K8s usage.


CathPag (approved):
Looks good. I don't have an opinion on the spelling. Following K8s makes sense .

Issue #19, opened 2021-02-09
Observability definition not started


JasonMorgan:
Need someone to fill in a definition for observability. You can find it [here](https://github.com/cncf/glossary/blob/main/definitions/observability.yaml)

PR #42, opened 2021-02-18
first pass at observability.md

IdealUsrname:
solves #19


CathPag:
I think this looks great! I kind of stumbled over this, though: "Observability is a **measure** of how a system's state..." Is it a measure or is it the ability measure? Also, feel like measure isn't the right word but can't think of a better one.

Full disclosure: I'm not too familiar with observability and this may just reflect my lack of knowledge.

For the last paragraph, I'd suggest the changes in bold:
The observability of a system has a significant impact on its cost. Observable systems yield meaningful, actionable data to their users, leading to more favorable outcomes and less downtime. **But note**, more information does not **necessarily** mean **your system** is more **observable. Decision**-makers should consider the tools and metrics that are feeding information for their **specific** decisions (or needs?).

IdealUsrname:
@CathPag How about

```Observability is a measurement of how well a system's state, or status, can be understood from its external outputs.```

Observability is typically defined as a measurement. I'd be wary of moving away from a previously defined term.

CathPag:
Ok, if that's the way it's generally described, that's how we should describe it :)

PR #41, opened 2021-02-18
First pass containerization.md

IdealUsrname:
solves #18 

I figure we should create a container image definition as well?

PR #45, opened 2021-02-18
Update containerization.md

CathPag:
A few stylistic changes. Feel free to accept or drop.


CathPag:
...and I got rid of the (see container images) as terms will link to one another. Created an issue so we don't forget to do that before we go live.

IdealUsrname (approved):

IdealUsrname:
good additions and changes to #41

Issue #43, opened 2021-02-18
Container Image - New Definition


IdealUsrname:
I was doing the first draft of the containerization definition. I realized we do not define a container image.

A first pass of a **Container Image** definition would be welcomed.

SamDelamarter:
I'll take a first swag at this. It doesn't look like I have the ability to assign the issue to myself.

CathPag:
@IdealUsrname I guess we can close this, right?

PR #46, opened 2021-02-19
Add initial definition of Container Image

SamDelamarter:
Solves #43


IdealUsrname (approved):
Made small changes to bring into line with other definitions.

Good first pass. We may need to align all the various "container definitions" in the future.

PR #56, opened 2021-03-04
Fixes #35 Remove duplicate stateful apps entry

### LINKED ISSUE #35: Two stateful app entries
Tom Leyden, could you please address this? You mentioned your stateful app entry wasn't in the glossary, so I added part of it per our email conversation. We did have one, though, and we have two: Stateful Apps [https://github.com/cncf/glossary/blob/main/definitions/stateful_apps.md](url) and Stateful vs Stateless Apps [https://github.com/cncf/glossary/blob/main/definitions/stateful_vs_stateless_applications](url) The latter is yours. 

Could you please add what's missing in the original entries and delete the one latter one?

SamDelamarter:
@CathPag It looks like all of the content @TomLeyden created in the Google doc was brought into the [stateful_apps.md](https://github.com/cncf/glossary/blob/main/definitions/stateful_apps.md) entry.

Let me know if I misunderstood what needs to be done here - I only deleted the duplicate file.


CathPag:
@SamDelamarter, yes, and now we have three entries:

1. stateful_apps.md
2. stateful_vs_stateless_applications.md
3. stateless_apps.md

We need one for stateful and another one for stateless. Don't want to delete the stateful vs stateless entry if there is good info in there. Wanna consolidate? Tom didn't have time, so if you'd like to tackle that, that'd be great!

JasonMorgan (approved):
LGTM

PR #67, opened 2021-03-12
CD V1-ready

CathPag:


JasonMorgan (approved):
LGTM, I expect this one will get a few more edits before we go live.

IdealUsrname (approved):
Are we adding links to the glossary items?

I'm a fan of a reference section where possible.

CathPag:
Good point. I prefer that too and we were going to add a further reading section in the future anyways. I'll update before the end of the week.

PR #71, opened 2021-03-12
Update devsecops.md

danpopSD:
suggested some additional definition.

Signed-off-by: danpopsd <dan.papandrea@sysdig.com>


JasonMorgan:
So this is going to overlap with the changes in #71. Also just being real those titles were mostly just me being sarcastic. Does #71 work for you? If so we can close this.

danpopSD:
do we have a "synonym" or may also be called?

JasonMorgan:
We don't right now but that's a good idea. Do you mind adding that once #70 has been merged?

danpopSD:
sounds good!

PR #106, opened 2021-04-21
Update contributing

CathPag:


JasonMorgan:
Fix the link and we should be good to go.

JasonMorgan (approved):

PR #115, opened 2021-05-04
chore: Ignore casing during spell checks

tomkerkhove:
Ignore casing during spell checks to reduce list of allowed words.


JasonMorgan (approved):
Nice!

PR #116, opened 2021-05-05
VMs make it easier to modify or replace HW

ArminMW:


JasonMorgan (approved):
Good addition. The whole definition will need to be reformatted but I think this is good content.

PR #118, opened 2021-05-05
abstraction create models

ArminMW:


JasonMorgan (changes requested):
Your addition didn't make the definition clearer or more accessible. Feel free to try rewording it but right now I don't understand what point you're making.

That being said good catch on the double period. Happy to have you reformat or remove the line from the first paragraph.

ArminMW:
requires clarification

PR #120, opened 2021-05-05
portability avoids lock-in

ArminMW:


ArminMW (commented):
Do these changes address the adaption to the style guide? Capital 1st letter for category, deletion of some hash chars?

CathPag:
@ArminMW, portability is a property, and because it describes a system's characteristic, the breakdown into "what it is," "problem it..." feels forced. That's why we decided to simply describe properties and skip the other sections. 

This is what the style guide states but I guess it's easily missed with so much info. We should probably think of a way to make it pop more:

"Three sections are required for terms in the technology and concept categories. Property definitions only require the first section."

JasonMorgan (approved):
Made a small change to the title and subtitles

PR #133, opened 2021-05-12
Style guide update

CathPag:
Tried to make the fact that "properties" don't need the last two sections a little more clear. Right now it's easy to miss.


caniszczyk (approved):
LGTM

PR #156, opened 2021-07-01
Bare metal V2

CathPag:


definitions/bare_metal_machine.md
```diff
@@ -1,17 +1,20 @@
...
+### The problem it addresses
 
-Bare metal gives you more control and computational power and is typically less expensive than VMs because you can purchase only the resources you need for the time that you need them. Application and web traffic can be “bursty” in nature  and often has heavy simultaneous usage from end users while also having long periods of low traffic or almost no usage.  When hosting in a cloud or hybrid cloud/ prem environment often you need to buy and configure resources to sustain the peak and high points of the traffic and usage but you also end up with long periods of over-configuration because of low traffic and this leads to higher cost for both hosting and maintenance.  
+Pairing one operating system with one physical computer is the original pattern of computing. All the resources of the physical computer are available directly to the operating system and with no virtualization layer present, there is no artificial delay in translating operating system instructions to hardware.
```
murillodigital, line 14: Additional problems bare metal addresses is data sovereignty (and compliance and governance in some cases), wonder if those would be valuable to add.

murillodigital (commented):
Added comment

definitions/bare_metal_machine.md
CathPag: Thanks, Leonardo. We'll need a little context here. I'll add it at the end of the paragraph. Could you finish the sentence? 

Pairing one operating system with one physical computer is the original pattern of computing. All the resources of the physical computer are available directly to the operating system and with no virtualization layer present, there is no artificial delay in translating operating system instructions to hardware. Additionally, bare metal helps address data sovereignty (and compliance and governance in some cases) by...

PR #167, opened 2021-07-13
Cluster V2

CathPag:


murillodigital (commented):
I have always understood clusters as the underlying compute nodes and associated infrastructure services (network, storage) that provide an abstracted layer for running and operating workloads (services) and not clusters as groups of applications (services) as is implied in this definition. Is that my own misunderstanding?

caniszczyk (approved):

Issue #158, opened 2021-07-01
Horizontal scaling definition


CathPag:
Please follow format defined in [style guide](https://github.com/cncf/glossary/blob/main/style-guide.md).

himanshu007-creator:
Hi i would like to work on this issue

CathPag:
Thanks, @himanshu007-creator. Once you've submitted a PR for the other term and it was accepted, we can assign you the next one. Hope that's ok.

sayantani11:
Hey I would like to contribute to this issue. Thank you.

CathPag:
Thanks for helping, @sayantani11! Please make sure to review the [style guide](https://github.com/cncf/glossary/blob/main/style-guide.md) before getting started.

sayantani11:
@CathPag I will.... Thanks for assigning

reddymahendra52:
Hello, Can I work on this issue too?
I would also like to contribute to this community.

CathPag:
@reddymahendra52, thanks for wanting to help! 
As you can see above @sayantani11 already claimed this term. Please pick one that hasn't been assigned yet.

PR #170, opened 2021-07-21
Horizontal Scaling definition

sayantani11:
Solves issue #158

### DIFF
--- a/definitions/horizontal_scaling.md
+++ b/definitions/horizontal_scaling.md
@@ -8,7 +8,7 @@ category: Concept
 
 ### What it is
 
-In [cloud computing](https://github.com/cncf/glossary/blob/main/definitions/cloud_computing.md), horizontal scaling is a technique where a system's capacity is increased by adding more [nodes](https://github.com/cncf/glossary/blob/main/definitions/nodes.md) versus adding more compute resources to individual nodes (the latter being known as vertical scaling). Let's say, we have a system of 4GB RAM and want to increase its capacity to 16GB RAM, scaling it horizontally means doing so by adding 4 x 4GB RAM rather than switching to a 16GB RAM system.
+Horizontal scaling is a technique where a system's capacity is increased by adding more [nodes](https://github.com/cncf/glossary/blob/main/definitions/nodes.md) versus adding more compute resources to individual nodes (the latter being known as vertical scaling). Let's say, we have a system of 4GB RAM and want to increase its capacity to 16GB RAM, scaling it horizontally means doing so by adding 4 x 4GB RAM rather than switching to a 16GB RAM system.
 
 This approach enhances the performance of an application by adding new instances, or [nodes](https://github.com/cncf/glossary/blob/main/definitions/nodes.md), to better distribute the workload. In simple words, it aims to decrease the server's load rather than expanding capacity of the individual server.
 
--- a/definitions/horizontal_scaling.md
+++ b/definitions/horizontal_scaling.md
@@ -7,17 +7,19 @@ category: Concept
 ## Horizontal Scaling
 
 ### What it is
+
 In [cloud computing](https://github.com/cncf/glossary/blob/main/definitions/cloud_computing.md) perspective Horizontal Scaling is type of [scalability](https://github.com/cncf/glossary/blob/main/definitions/scalability.md) whose approach is to add additional instances instead of moving into a larger instance size.
 
 For example, let us consider we have a system of 4GB RAM and we want to increase it's capacity to 16GB RAM, Horizontal Scaling means doing this by making a system of 4 X 4GB RAM rather than switching to a 16GB RAM system.
 
-It is an approach to enhance the performance of the server [node](https://github.com/cncf/glossary/blob/main/definitions/nodes.md) by adding new instances of the server to already existing servers in order to distribute the workload equally.
-In simple words it aims to decrease the server's load rather than expanding capacity of the individual server.
+It is an approach to enhance the performance of the server [node](https://github.com/cncf/glossary/blob/main/definitions/nodes.md) by adding new instances of the server to already existing servers in order to distribute the workload equally. In simple words it aims to decrease the server's load rather than expanding capacity of the individual server.
 
 Load balancing, clustering, and distributed file systems are some of the crucial and popular strategies of horizontal scaling. It caters to portioning of data where each [node]((https://github.com/cncf/glossary/blob/main/definitions/nodes.md)) contains single part of the data.
 
 ### Problem it addresses
+
 With growing demands of the application the system should be able to scale by increasing/growing/managing the resources for the customers using it. The latency of the API calls might increase due to heavy traffic, this is the problem that can be solved by horizontal scaling.
 
 ### How it helps
+
 Horizontal scaling helps as it can provide limitless scaling with endless additions to server instances. It enhances the I/O concurrency, and also reduces the load on existing [nodes](https://github.com/cncf/glossary/blob/main/definitions/nodes.md) and expands disk capacity.
--- a/definitions/horizontal_scaling.md
+++ b/definitions/horizontal_scaling.md
@@ -7,13 +7,17 @@ category: Concept
 ## Horizontal Scaling
 
 ### What it is
-In cloud computing perspective Horizonatal Scaling means adding additional instances instead of moving into a larger instance size. It is an approach to enhance the perfomance of the server node by adding new instances of the server to already existing servers in order to distribute the workload equally. It basically aims to decrease the server's load rather than expanding capacity of the individual server.
+In [cloud computing](https://github.com/cncf/glossary/blob/main/definitions/cloud_computing.md) perspective Horizontal Scaling is type of [scalability](https://github.com/cncf/glossary/blob/main/definitions/scalability.md) whose approach is to add additional instances instead of moving into a larger instance size.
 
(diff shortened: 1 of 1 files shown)


sayantani11:
Please let me know if any changes are required

CathPag:
Thanks, @sayantani11! Some of our maintainers are on vacation, so please allow us a little time to review and get back to you!

CathPag:
Thanks for your submission. This will need some refinement but is a good start. Here's some feedback:

- Please make sure to spell check before submitting. There are quite a few typoes. 
- Keep in mind that you may be explaining concepts to someone with no technical background. "In cloud computing perspective Horizontal Scaling means adding additional instances instead of moving into a larger instance size," doens't bring a lot of clarity. What does adding instances and moving to a larger instance mean? Explaining something well is not easy and takes time. Also, consider referencing the node definition 
- If appropriate, try to use an example that everyone knows

Please make sure to read the [style guide](https://github.com/cncf/glossary/blob/main/style-guide.md) (we added a section at the bottom). Hopefully, that will give you a better sense of what we need.

sayantani11:
Thanks for the review @CathPag 
I am looking into it.

JasonMorgan:
@sayantani11 The changes you made to the What it is section are amazing! It's very close to being ready to publish. That being said the problem it addresses and how it helps needs work. I'll make some more comments in line.

definitions/horizontal_scaling.md
```diff
@@ -0,0 +1,25 @@
...
+
+For example, let us consider we have a system of 4GB RAM and we want to increase it's capacity to 16GB RAM, Horizontal Scaling means doing this by making a system of 4 X 4GB RAM rather than switching to a 16GB RAM system.
+
+It is an approach to enhance the performance of the server [node](https://github.com/cncf/glossary/blob/main/definitions/nodes.md) by adding new instances of the server to already existing servers in order to distribute the workload equally. In simple words it aims to decrease the server's load rather than expanding capacity of the individual server.
+
+Load balancing, clustering, and distributed file systems are some of the crucial and popular strategies of horizontal scaling. It caters to portioning of data where each [node]((https://github.com/cncf/glossary/blob/main/definitions/nodes.md)) contains single part of the data.
+
+### Problem it addresses
+
+With growing demands of the application the system should be able to scale by increasing/growing/managing the resources for the customers using it. The latency of the API calls might increase due to heavy traffic, this is the problem that can be solved by horizontal scaling.
```
JasonMorgan, line 21: "increasing/growing/managing" we can't have the slashes in there. Ideally turn this into a couple complete sentences that convey the key problem horizontal scaling is meant to address. Something like: 

As demand for an application grows beyond the current capacity of that application instance an operator needs to find a way to add capacity by scaling. Either by adding more application instances or by giving more compute resources to the existing instances, or said another way, by scaling your application horizontally or vertically.

definitions/horizontal_scaling.md
```diff
@@ -0,0 +1,25 @@
...
+
+### How it helps
+
+Horizontal scaling helps as it can provide limitless scaling with endless additions to server instances. It enhances the I/O concurrency, and also reduces the load on existing [nodes](https://github.com/cncf/glossary/blob/main/definitions/nodes.md) and expands disk capacity.
```
JasonMorgan, line 25: So at a minimun we don't want to use any abbreviations so you'd need to spell out I/O. Even better consider that the audience for this is folks that don't necessarily know a ton of computer science terms and instead use language that conveys your meaning in a more accessible way. Example:

Horizontal scaling allows an application to process a greater number of requests by creating more nodes, or application instances, that can perform the work. If a single node can handle 1000 requests per second then each additional node should increase the total number of requests by around 1000 requests per second.

JasonMorgan (changes requested):

definitions/horizontal_scaling.md
sayantani11: I get it that it is folks who would read to understand. From next time I would try to keep it in a way that anyone would could understand and as simple as possible

definitions/horizontal_scaling.md
```diff
@@ -0,0 +1,25 @@
...
+title: Horizontal Scaling
+status: Feedback Appreciated
+category: Concept
+---
+
+## Horizontal Scaling
+
+### What it is
+
+In [cloud computing](https://github.com/cncf/glossary/blob/main/definitions/cloud_computing.md) perspective Horizontal Scaling is type of [scalability](https://github.com/cncf/glossary/blob/main/definitions/scalability.md) whose approach is to add additional instances instead of moving into a larger instance size.
```
CathPag, line 11: This is great. A few points:

- Not everyone may be familiar with what instance means in this context. Node is a term that has already been captured in the glossary, so, if we use node instead of instance, people can easily look it up
- You say "adding additional instances," but to what? The subject of the sentence is missing 
-  Since the example refers to the sentence before, it should be part of the same paragraph 
- Horizontal scaling is a qualified noun, not a name, and should not be capitalized

Here's my suggestion: 

Horizontal scaling is a scaling technique where a system's capacity is increased by adding more [nodes](https://github.com/cncf/glossary/blob/main/definitions/nodes.md) versus adding more compute resources to individual nodes (the latter is referred as vertical scaling). For example, ...

definitions/horizontal_scaling.md
```diff
@@ -0,0 +1,25 @@
...
+
+For example, let us consider we have a system of 4GB RAM and we want to increase it's capacity to 16GB RAM, Horizontal Scaling means doing this by making a system of 4 X 4GB RAM rather than switching to a 16GB RAM system.
+
+It is an approach to enhance the performance of the server [node](https://github.com/cncf/glossary/blob/main/definitions/nodes.md) by adding new instances of the server to already existing servers in order to distribute the workload equally. In simple words it aims to decrease the server's load rather than expanding capacity of the individual server.
```
CathPag, line 15: I believe "server node" isn't the right description here. Nodes run on a server and you may add more nodes on the same server (@JasonMorgan, correct me if I'm wrong) . Here's my suggested change:

This approach enhances system performance by adding additional instances, or nodes, to an existing server and distributing the workload equally among them. In simple words, horizontal scaling decreases the instance's load rather than expanding its capacity.

As you can see, I introduced "instance" but also refer back to nodes so the reader understands it's the same thing.

definitions/horizontal_scaling.md
```diff
@@ -0,0 +1,25 @@
...
+category: Concept
+---
+
+## Horizontal Scaling
+
+### What it is
+
+In [cloud computing](https://github.com/cncf/glossary/blob/main/definitions/cloud_computing.md) perspective Horizontal Scaling is type of [scalability](https://github.com/cncf/glossary/blob/main/definitions/scalability.md) whose approach is to add additional instances instead of moving into a larger instance size.
+
+For example, let us consider we have a system of 4GB RAM and we want to increase it's capacity to 16GB RAM, Horizontal Scaling means doing this by making a system of 4 X 4GB RAM rather than switching to a 16GB RAM system.
```
CathPag, line 13: Great example!

CathPag (changes requested):
Looks great. Offered a few suggestions and explained why I feel they are necessary. Hope it makes sense.

definitions/horizontal_scaling.md
JasonMorgan: I'm going to make an edit in line to clean this up. I'm glad we got the definition  of node in here but the usage is a little confusing.

definitions/horizontal_scaling.md
JasonMorgan: Awesome! Do you want to make the changes?

definitions/horizontal_scaling.md
```diff
@@ -0,0 +1,25 @@
...
+
+It is an approach to enhance the performance of the server [node](https://github.com/cncf/glossary/blob/main/definitions/nodes.md) by adding new instances of the server to already existing servers in order to distribute the workload equally. In simple words it aims to decrease the server's load rather than expanding capacity of the individual server.
+
+Load balancing, clustering, and distributed file systems are some of the crucial and popular strategies of horizontal scaling. It caters to portioning of data where each [node]((https://github.com/cncf/glossary/blob/main/definitions/nodes.md)) contains single part of the data.
```
JasonMorgan: I think it would be best to remove this line. It gets too into the details.

definitions/horizontal_scaling.md
sayantani11: @JasonMorgan Ofcourse why not

definitions/horizontal_scaling.md
sayantani11: @JasonMorgan Yeah sure I also had this thought, as it aims to give a basic idea to the person reading. I felt this was a bit too much of info for someone starting. Apart from that the terms may also be difficult for a newbie to understand.
I was unsure myself while adding it.

CathPag (approved):
We just tweaked a bit the wording but other than that, this looks great! 
Btw, we removed "in cloud computing" because horizontal scaling doesn't only apply to cloud computing.

JasonMorgan:
@sayantani11 we made some changes to your definition and marked the entry as completed and ready to go in the next glossary release. Do you mind reviewing the entry and giving me a thumbs up if you're happy with it?

sayantani11:
Thanks for the help @CathPag and @JasonMorgan

JasonMorgan (approved):

PR #175, opened 2021-07-23
Nodes updated to new format

Utkarshn10:
References: https://kubernetes.io/docs/concepts/architecture/nodes/#:~:text=Node%20conditions%2C%20and%20a%20description%20of%20when%20each,the%20processes%20...%20%201%20more%20rows%20
https://www.oracle.com/cloud-native/container-engine-kubernetes/what-is-kubernetes/#link1

### DIFF
--- a/definitions/nodes.md
+++ b/definitions/nodes.md
@@ -1,13 +1,13 @@
 ---
 title: Nodes
 status: Feedback Appreciated
-category: concept
+category: Concept
 ---
 ## Nodes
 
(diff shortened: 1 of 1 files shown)


JasonMorgan:
@Utkarshn10 going to reject these changes. The intent of the definition is not to be Kubernetes specific as a node could apply to any number of clustered systems. Additionally, we're avoiding terms like master for the control plane or leader components in a cluster. I appreciate the attempt but this change moves the glossary in the wrong direction.

PR #181, opened 2021-08-01
Updated Virtual Machines to New Format

TakshakRamteke:
### Changes I have made :

- [x] Updated the entity to new format.
- [x] Checked the definition for typos/grammatical mistakes.
- [x] I have also asked some of my friends from Technical and Non-Technical backgrounds if they were able to understand the all the sections. To which they responded positively.

I have not made a whole lot of changes to the previously written content although I have referred some of the complete definition and websites for the technical soundness of the statements but I still am a little confused if the definition and other things are up to the mark.

Also on the definition by **redHat organization** i came across a term **Hypervisor** which is not present in this project which but is an integral software for creating virtual machines so i think it would be nice if we could add it to glossary as well.

### Feedback Appreciated
_Please let me know if i need to make any further changes to this same entity (i.e. virtual machines)._

### DIFF
--- a/definitions/virtual_machine.md
+++ b/definitions/virtual_machine.md
@@ -1,13 +1,15 @@
 ---
 title: Virtual Machine
 status: Feedback Appreciated
-category: Software
+category: Technology
 ---
 ## Virtual Machine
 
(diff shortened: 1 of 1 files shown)


CathPag:
Thank you, @TakshakRamteke. As mentioned to other contributors, we are currently understaffed and have to work our way through the 20 PRs that have been submitted. Please give us some time to review and provide feedback.

JasonMorgan (approved):
Awesome! Thank you!

CathPag (approved):
LGTM, Thank you!

PR #209, opened 2021-08-19
Updated Tightly coupled Architecture's

TakshakRamteke:
### Changes I Made 
- [x] Updated the problem it addresses section.
- [x] Checked the entity and for Typos & Grammatical mistakes and made the changes accordingly.
- [x] Asked my friends and peers from technical and Non-Technical background if they were able to understand all the sections, To which i had a positive response.

I've made only a few necessary change to the definition section. As for the Problem it addresses section I've almost rewritten the entire section because it was addressing the disadvantage of another term.

- Before it was like this
```
A major disadvantage of Loosely Coupled Architecture is that it has a very low data rate and is very inconsistent.
```
- After the changes it looks something like this
```
Tightly Coupled Architecture's mainly addresses problems related to having a very low data rate and inconsistency.
```
As I wasn't able to find a clear answer as to  what are the problems addressed by Tightly coupled architecture


TakshakRamteke:
Please let me know if i any changes are required

CathPag:
Thanks for your suggestions, @TakshakRamteke, but it needed a lot more work so we rewrote it. Btw, tightly coupled architecture is a property and doesn't require the three sections per style guide. We do appreciate the PR and patience!

TakshakRamteke:
Thank you Would try to do better next time

PR #251, opened 2021-11-16
Create LOCALIZATION.MD

CathPag:
@murillodigital @JasonMorgan what else should go in here?

PR #253, opened 2021-11-18
Add Korean l10n for testing a new l10n

seokho-son:
This PR is to test a new localization. 

As an example, I added Korean localization.

I think we can check netlify preview of this PR and discuss to enhance localization guideline and enhancement points for this site. (ref: #251)

To add a new localization, we need to include
 - `[languages.ko]` section in `config.toml`
 - content/**ko**/ directory
 - Translated document for Home (content/ko/_index.md)

I hope to note that
- this PR includes a document translated into Korean. (it is also an example)
- there is no Korean localization team yet. 
- I am aware of this PR will not be approved :) 
- https://gohugo.io/functions/lang.merge/ seems not working. 
  - If we don't include translated file, that file is not visible in sidebar-tree

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -66,6 +66,14 @@ weight = 1
 #time_format_default = "02.01.2006"
 #time_format_blog = "02.01.2006"
 
+[languages.ko]
+title = "클라우드 네이티브(Cloud Native) 용어집"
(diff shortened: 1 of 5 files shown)


seokho-son:
Let me close this PR :)
PR #291 is the modified version of this PR.

PR #259, opened 2021-11-24
Update stateless_apps.md

JatinNarang1:
@CathPag I've tried to simplify the the definition of stateless application, kindly review it. :)

### DIFF
--- a/content/en/stateless_apps.md
+++ b/content/en/stateless_apps.md
@@ -6,7 +6,7 @@ category: technology
 
 ## What it is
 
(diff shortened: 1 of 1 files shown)


CathPag:
@JatinNarang1, as mentioned in the issue, we want to mirror the [stateful definition ](https://github.com/cncf/glossary/blob/main/content/en/stateful_apps.md). Have a look at [horizontal](https://glossary.cncf.io/horizontal_scaling/) and [vertical scaling](https://glossary.cncf.io/vertical_scaling/). See how they work well together? The reasoning is that people will likely read both together and they should work as a "couple." Does that make sense?

CathPag:
@JatinNarang1, since we haven't heard from you, I'll close this PR and open it up for other people to work on it

Issue #280, opened 2021-12-17
Create an entry for data center


JasonMorgan:

kaihoffman:
I'll take a stab at this.

PR #282, opened 2021-12-17
Create entry for "data center"

kaihoffman:
Add proposal for a definition of data center (issue #280 )

### DIFF
--- a/content/en/data_center.md
+++ b/content/en/data_center.md
@@ -1,0 +1,17 @@
+---
+title: Data center
+status: Feedback Appreciated
+category: Technology
+---
+
+## What it is
+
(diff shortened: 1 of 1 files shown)


CathPag (changes requested):
This is a great entry, @kaihoffman! Only one little tweak, please. 

In the problem it addresses, you are touching upon how it helps (the next section): data centers allow companies to..."  This section should fully focus on the problem statement which you mentioned in the 2nd sentence. 

Could you rephrase this so:

- the problem it addresses only focuses on the problem
- move what you say in the first sentence to the next section

Thank you!

CathPag (approved):
Approved as feedback appreciated but we still need to weave in this feedback:

In the problem it addresses, you are touching upon how it helps (the next section): data centers allow companies to..." This section should fully focus on the problem statement which you mentioned in the 2nd sentence.

Could you rephrase this so:

- the problem it addresses only focuses on the problem
-  move what you say in the first sentence to the next section

PR #296, opened 2022-01-06
Fix configuration for branch protection

seokho-son:
The current settings for branch protection seems not work properly.
(I think some settings of https://github.com/apps/settings are not stable yet)

This PR updates `.github/settings.yml` and `CODEOWNERS` to fix configuration on branch protection.
I also suggest following configuration for permission.
- All maintainers (admins) have push permission to all branches.
- Approvers for each localization team have push permission to corresponding development branch (e.g. `dev-ko`)
  - but they cannot push a commit for files not related with localization contents unless an appropriate owner(reviewer) in CODEOWNERS gives an approving review.
  - they cannot push a commit to `main` branch

I'v tested `https://github.com/apps/settings` with a similar configuration in another organization :) 
I hope this PR resolves all :)

Please note that
- I temporally set myself as one of `admin` to check this setting by `https://github.com/apps/settings` works properly. (we can check this from `setting` tap)
- I intentionally removed branch setting for Portuguese temporally. (Let's check configuration for Korean branch works firstly)

PR #299, opened 2022-01-09
Update settings.yml for dev-hi branch protection and approvers

anubha-v-ardhan:
This PR updates settings.yml with dev-hi branch protection rule and approvers.

### DIFF
--- a/.github/settings.yml
+++ b/.github/settings.yml
@@ -49,5 +49,15 @@ branches:
         users:
          - seokho-son
          - Eviekim
-         - jihoon-seo
- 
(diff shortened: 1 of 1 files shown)


seokho-son:
@anubha-v-ardhan The current setting for dev-branch does not work properly. 
I think this PR would not work properly too. 
We still need to check permission issue. (at least PR #296 needs to be merged and tested)
Let me close this PR to prevent accidental approvals :)

anubha-v-ardhan:
Oh!, no issues let us wait for #296 then

PR #302, opened 2022-01-12
Add GitOps definition draft for submission

hseligson1:
I'm taking a stab at this and I am open to any/all feedback, especially a way to simplify the "The Problem it Addresses" section. Looking forward to your comments!

### DIFF
--- a/content/en/gitops.md
+++ b/content/en/gitops.md
@@ -1,0 +1,40 @@
+---
+title: GitOps
+status: Feedback Needed
+category: Concept
+---
+
+## What it is
+
+GitOps is a set of best practices based on [shared principles](https://opengitops.dev/) applied to managing and operating a cloud-native infrastructure and applications. 
+The GitOps approach is used to strategize a cohesive workflow that leverages Git as the "single source of truth" and extends this practice to an application's configuration, infrastructure, and operational procedures. 
+The GitOps controller automates matching the Git state with the system state in a bidirectional manner.
+
+## Problem it addresses
+
+Some existing processes for infrastructure configuration management can face challenges, such as:
+
+- Failed deployments
+
+Relying on a disaster recovery strategy with manual or semi-manual processes isn't always reliable. 
+
+- Configuration drift
+
+This approach can create consequences, such as security vulnerabilities, high costs due to lack of efficiencies resulting in stale workloads, and risking downtime in an application or service because it's not scaling enough or is scaling too frequently.
+
+- Poor Infrastructure design
+
+Legacy infrastructure designs limit developers to access or understand large-scale systems that limit their productivity.
+
+- Missing documentation or unknown development history
+
+Many teams depend on working practices developed over decades and document changes in a centrally maintained wiki that is uncontrolled and not maintained.
+
+Adopting a GitOps workflow can help alleviate some of these issues.
+
+## How it helps
+
+GitOps is a deployment practice that improves the management of a cloud system infrastructure. It helps by enabling a declarative format that a team can describe how their infrastructure works. It also helps describe the entire system running on top of it. This approach provides traceability to any changes that are made within any environment and supports automated rollbacks, self-healing attributes, and easier recovery during service outages.
+By operating in a close loop it ensures that the desired state of a system matches a declared state by utilizing software agents to protect against human error or manual intervention.
+
+GitOps can increase a team's productivity, allowing them to deploy more frequently, automate rollbacks and remediate outages quickly and effectively.
--- a/content/en/gitops.md
+++ b/content/en/gitops.md
@@ -1,6 +1,6 @@
 ---
 title: GitOps
-status: Feedback Needed
+status: Feedback Appreciated
 category: Concept
 ---
 
--- a/content/en/gitops.md
+++ b/content/en/gitops.md
@@ -6,35 +6,15 @@ category: Concept
 
 ## What it is
 
-GitOps is a set of best practices based on [shared principles](https://opengitops.dev/) applied to managing and operating a cloud-native infrastructure and applications. 
-The GitOps approach is used to strategize a cohesive workflow that leverages Git as the "single source of truth" and extends this practice to an application's configuration, infrastructure, and operational procedures. 
-The GitOps controller automates matching the Git state with the system state in a bidirectional manner.
+GitOps is a set of best practices based on [shared principles](https://opengitops.dev/), applied to a workflow that depends on software tools that enable automation to reconcile a declared system state or configuration in a git repository.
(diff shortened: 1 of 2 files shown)


content/en/gitops.md
```diff
@@ -0,0 +1,40 @@
...
+title: GitOps
+status: Feedback Appreciated
+category: Concept
+---
+
+## What it is
+
+GitOps is a set of best practices based on [shared principles](https://opengitops.dev/) applied to managing and operating a cloud-native infrastructure and applications. 
+The GitOps approach is used to strategize a cohesive workflow that leverages Git as the "single source of truth" and extends this practice to an application's configuration, infrastructure, and operational procedures. 
+The GitOps controller automates matching the Git state with the system state in a bidirectional manner.
```
ChrisJBurns, line 11: Instead of `controller` possibly the better work is `agent`? or `operator`? 

Ref: https://www.weave.works/technologies/gitops/

content/en/gitops.md
```diff
@@ -0,0 +1,40 @@
...
+
+Legacy infrastructure designs limit developers to access or understand large-scale systems that limit their productivity.
+
+- Missing documentation or unknown development history
+
+Many teams depend on working practices developed over decades and document changes in a centrally maintained wiki that is uncontrolled and not maintained.
+
+Adopting a GitOps workflow can help alleviate some of these issues.
+
+## How it helps
```
ChrisJBurns, line 35: Another reason is to enhance the security posture of an organisation. Using GitOps, organisations subscribe to a `pull` based approach as opposed to the `push` based approach that has more commonly been followed. Therefore, if a CI system (Jenkins etc) is compromised, the attack surface is now smaller because CI and CD are separated - as traditionally people deploy to Kubernetes via their CI systems using push strategies.

content/en/gitops.md
hseligson1: @ChrisJBurns - Hm, so something like "GitOps depends on software agents that automate matching the Git state with a system state in a bidirectional matter."? Because it wouldn't be "GitOps agent" or "GitOps operator".

content/en/gitops.md
hseligson1: @ChrisJBurns - I hesitate to focus on a pull/push-based operation approach only and mention Jenkins because this limits the context to remain vast for the audience who utilize other tools and approaches for their GitOps workflow.

content/en/gitops.md
```diff
@@ -0,0 +1,40 @@
+---
+title: GitOps
+status: Feedback Appreciated
+category: Concept
+---
+
+## What it is
+
+GitOps is a set of best practices based on [shared principles](https://opengitops.dev/) applied to managing and operating a cloud-native infrastructure and applications. 
+The GitOps approach is used to strategize a cohesive workflow that leverages Git as the "single source of truth" and extends this practice to an application's configuration, infrastructure, and operational procedures.
```
JasonMorgan, line 10: Can we rephrase this?

content/en/gitops.md
JasonMorgan: @hseligson1 really great start here but can we try and bring the entire problem section down to 3-5 lines.

content/en/gitops.md
```diff
@@ -0,0 +1,40 @@
...
+Many teams depend on working practices developed over decades and document changes in a centrally maintained wiki that is uncontrolled and not maintained.
+
+Adopting a GitOps workflow can help alleviate some of these issues.
+
+## How it helps
+
+GitOps is a deployment practice that improves the management of a cloud system infrastructure. It helps by enabling a declarative format that a team can describe how their infrastructure works. It also helps describe the entire system running on top of it. This approach provides traceability to any changes that are made within any environment and supports automated rollbacks, self-healing attributes, and easier recovery during service outages.
+By operating in a close loop it ensures that the desired state of a system matches a declared state by utilizing software agents to protect against human error or manual intervention.
+
+GitOps can increase a team's productivity, allowing them to deploy more frequently, automate rollbacks and remediate outages quickly and effectively.
```
JasonMorgan, line 40: Same as my last comment, we can pair this whole section down to 3-5 lines.

JasonMorgan (changes requested):

content/en/gitops.md
ChrisJBurns: How about something like:

_"GitOps depends on software agents that aim to automatically reconcile declared state in Git with the realised state in system infrastructure"_ ?

content/en/gitops.md
ChrisJBurns: That's a fair point!

content/en/gitops.md
hseligson1: @ChrisJBurns - I like that! I'll submit that change

content/en/gitops.md
hseligson1: Yes - I'll submit a simplified version

content/en/gitops.md
hseligson1: @ChrisJBurns - I ended up trying to simplify this as much as I could and make it more general. When I shared the term "agent" or "controller" with non-technical people they were unsure what that was. I included tools instead, but curious what you and @JasonMorgan think

ChrisJBurns: I'm not sure using the word `tools` would be a good fit here. Tools are used to complete a task whereas Agents do the task for you - following [this article](https://medium.com/@dieswaytoofast/ftools-vs-agents-know-your-product-edition-2a2d284b3263). Although I understand it's a tricky one for non-technical folk, we want to be careful when we avoid the use of technical terms for non-technical folk benefit, that we don't actually use words in replacement that mislead the definition itself.

hseligson1: @ChrisJBurns - This is a great point, "Tools are used to complete a task whereas Agents do the task for you". I also really enjoyed [MIT's](https://www.media.mit.edu/groups/software-agents/overview/) definition and will try to think of a simple way to describe this, otherwise, I'll default to "agent".

hseligson1: This was updated!

content/en/gitops.md
hseligson1: @JasonMorgan - this was updated and revised

JasonMorgan: Got it, I'll look through now.

CathPag (approved):
Looks good for "feedback appreciated" as it needs to be simplified

JasonMorgan (approved):
We've got enough here to accept at Feedback Appreciated. Thanks for your work @hseligson1, this is a great start to a particularly difficult term.

JasonMorgan:
@hseligson1 Can you appease the spellchecker?

hseligson1:
> 

Will do!

Issue #269, opened 2021-12-03
Tracking Portuguese Localization Progress


edsoncelio:
Issue to track Portuguese localization progress.

## Checklist to translate a term
1. Check that no one else has claimed the term in the comments below.
2. Comment below the term that you would like to translate, focus on one term per PR.
3. Fork this repo, make your translation and open a PR with the following name convention: `[Add/Update/Fix] [term path]`, for example:
`Add content/pt/abstraction.md`

For more details and any information that you need, please don't hesitate and ask on the [Portuguese slack channel](https://cloud-native.slack.com/archives/C02MWQTJUPL).

jessicalins:
Hello :) 

I'd like to translate `content/pt/application_programming_interface.md` and `content/pt/blue_green_deployment.md`. I'll open PRs for them shortly.

I've tried to translate the content in `pt-BR` - I was wondering if we want to differentiate `pt-pt` from `pt-BR` or for the sake of simplicity we consider both as `pt`? What do you think?

edsoncelio:
Hello, thanks @jessicalins. For now, most of the team members are Brazilians, so I think is better to review in `pt-BR`, later if we found someone to help in `pt-pt` we can discuss more that, what do you think?

Ps.: I will assign to you the pages that you are working on, thanks again!

jessicalins:
@edsoncelio sounds good!

willlsantos:
Hi folks, are you ok?

I'd like to translante content/pt/api_gateway.md and content/pt/auto_scaling.md.

edsoncelio:
Thanks @willlsantos, I'll assign it to you if you need some help, just ping us on slack :)

MatheusPaes:
Hello,

I'll tranlsate content/pt/abstraction.md and content/pt/agile_software_development.md
The PR's will be open soon ;)

edsoncelio:
Thanks @MatheusPaes, if you need some help, just ping us on the slack channel :)

willlsantos:
Hello, 
I'd like to translate content/pt/client_server_architecture.md.

Adi9235:
Hello, @edsoncelio I am new to contributing to cncf projects in this issue we have to translate one term to Portuguese if I am right can I work on content/pt/auto_scaling.md.
Thank You

marcelomansur:
Hi @edsoncelio , I'd like to work on `content/pt-br/devops.md`.

MrErlison:
[...]

PR #313, opened 2022-01-14
Add content/pt/auto_scaling.md

MatheusPaes:
Translation auto_scaling.md to Portuguese. #269


edsoncelio (changes requested):
Hello @MatheusPaes, a few points about your PR:

* Could you update your branch with the `dev-pt` branch? Now we are using the content dir as `pt-br/` instead of `pt/`
* Could you submit just one definition per PR? To this for example we have 3 definitions in the same PR
* The header `category` can be translated, you can see [here](https://github.com/cncf/glossary/blob/dev-pt/content/pt-br/style-guide/_index.md) the values

And thanks!

PR #360, opened 2022-01-28
Enable non-english dictionaries

electrocucaracha:
[PySpelling](https://facelessuser.github.io/pyspelling/) tool can be configured to validate another languages, for example spanish.

This change moves the English spellchecker configuration into its own content folder. This approach allows to isolate the spellchecker execution by language or branch.

It's possible to validate this change locally thru the [act tool](https://github.com/nektos/act) running the `act -j check-english-spellcheck` command.

> Note: Some words were removed from the English wordlist


JasonMorgan:
Hey @electrocucaracha thanks for taking the time to chat with me today! As per our conversation please add some documentation explaining how to add support for a new language as well as documentation covering running the spellchecker locally. With that I should be able to get this merged by the end of the week. cc @seokho-son

.github/workflows/spellcheck.yml
```diff
@@ -1,28 +1,34 @@
...
-    branches: [ main ]
-
-  # Allows you to run this workflow manually from the Actions tab
+    types:
+      - submitted
+    paths:
+      - '**.md'
+      - '!.github/*'
+    branches:
+      - main
```
seokho-son, line 21: Each localization team has (/will have) their own development branch (https://github.com/cncf/glossary/tree/dev-ko for Korean for instance). So, localization teams will not use `main` branch for localizing documents. We need to add development branch if a localization team wants to use this spellchecker in workflow.

seokho-son (commented):
@electrocucaracha Thank you for the contribution to improve workflow :)

First of all, I need to share following situation.

 - Each localization team supposed to use their own development branch (ex: dev-pt)
 - There is no development branch for Spanish team yet (no request for Spanish team initiation yet)
 - This PR includes files in content/es/*

So, I think the spellchecker configuration for Spanish content would be better to be managed (configured) after a Spanish team is established with their own branch.
@JasonMorgan @CathPag what do you think so?

electrocucaracha:
> 
>     * Each localization team supposed to use their own development branch (ex: dev-pt)
> 

@seokho-son another approach could be the creation of GitHub workflows per language. This requires multiple spellcheck configuration files but the major benefit is granular control. I can change this PR to reflect that idea.

JasonMorgan:
> 

Based on requirements we have around permissions we already need to do unique branches for each language. Happy to try whatever approach as long as the spellchecks can safely run on their respective branches.

electrocucaracha:
I have created this [PR](https://github.com/cncf/glossary/pull/388) for the Spanish localization team where includes the changes required by the spellchecker. The drawback of using the `rojopolis/spellcheck-github-actions` is relying on the addition of dictionaries.

PR #371, opened 2022-01-31
Update Client-Server Architecture to provide example of desktop application and removing wording that suggests that desktop applications came before client-server architectures

dinoboy197:
The current wording in the client-server architecture document somewhat implies that self-contained applications came before client-server applications and that they are no longer in use (via the use of the word "legacy").

Being sensitive to this glossary focusing on CNCF technologies, I removed the word "legacy" and replaced it with a slightly more elaborate description of a self-contained application (desktop app) to clearly distinguish client-server from self-contained apps.

### DIFF
--- a/content/en/client_server_architecture.md
+++ b/content/en/client_server_architecture.md
@@ -6,7 +6,9 @@ category: concept
 
 ## What it is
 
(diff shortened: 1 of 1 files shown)


CathPag (approved):
Looks great, thank you!

JasonMorgan (approved):

JasonMorgan:
@dinoboy197 this is wonderful, thank you!

Issue #367, opened 2022-01-29
[English]: Golden Signals 


murriel:
### Details (optional)

Optional details

PadminiRai:
Can I work on it ?

CathPag:
Here you go, @PadminiRai. Please make sure to read the [style guide](https://glossary.cncf.io/style-guide/) before getting started and let us know if you have any questions.

JasonMorgan:
@remote007 how's it going? Is there anything we can do to help this move forward?

remote007:
Hi @JasonMorgan , I edited the content it's in the draft format rn. I fell ill and am taking some rest. Will create asap. Thanks for asking.

CathPag:
Apologies for the delay, but this term no longer meets the requirements to be included in the cloud native glossary. I will go ahead and close this issue.

PR #413, opened 2022-02-09
Definition for Golden Signals

remote007:
[English]: Golden Signals 
Fixes : https://github.com/cncf/glossary/issues/367


content/en/golden_signals.md
```diff
@@ -0,0 +1,22 @@
+---
+title: Golden Signals
+status: Feedback Appreciated
+category: concept
+---
+
+## What it is
+Golden signals contain a set of criteria that provide a wide view of a service from a user’s perspective . It is the gold standard for monitoring a web application’s performance.
```
CathPag, line 8: The first sentence is a mouth full :) Remember that we are trying to make this accessible to anyone.  

The most basic information we need to convey is that golden signals are metrics. Signal doesn't necessarily imply that. That's the first aha moment. Then we need to explain what kind of metrics. There are typically four: latency, traffic, errors, and saturation. Followed by a quick note as to why they are important. Why should you care about golden signals? Does that make sense?

Golden signals are key metrics for any application, not only web apps, btw.

content/en/golden_signals.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Problem it addresses
+Old monitoring tools are much suitable as these were created to monitor machines, not microservices. On the other hand, most of the modern monitoring systems offer a huge variety of parameters for various purposes, making it easy to drown in and lose focus of what really is relevant for the application.
```
CathPag, line 11: I think we should focus on the problem that occurs if we don't have that type of data. Without having latency, traffic, errors, and saturation data, operators are flying blind. What does typically goes wrong when you don't have that data?

CathPag (changes requested):
Great start, @remote007! Added two comments.

content/en/golden_signals.md
JasonMorgan: Also likely worth linking out to one of the google SRE docs where they define golden metrics, ex: https://sre.google/sre-book/monitoring-distributed-systems/#xref_monitoring_golden-signals

remote007:
@CathPag @JasonMorgan Thanks for the feedback , I'll start working on it.

content/en/golden_signals.md
```diff
@@ -0,0 +1,24 @@
+---
+title: Golden Signals
+status: Feedback Appreciated
+category: concept
+---
+
+## What it is
+[Golden signals](https://sre.google/sre-book/monitoring-distributed-systems/#xref_monitoring_golden-signals) are metrics that help us understand how [services](https://glossary.cncf.io/service/) are experienced by users. Today, it is the gold standard for monitoring application performance and focuses on four specific metrics: Latency, traffic, errors, and saturation. The first three metrics relate to requests. A request is a message sent from one service to another. Services do that to collaborate and work as a [distributed app](https://glossary.cncf.io/distributed_apps/).
+
+*Latency* is the time it takes to service a request. *Traffic* measures the request frequency placed on the application. *Errors* measure the request failure rate. And *Saturation* measures the capacity of app resources, allowing users to address bottlenecks before they become a problem.
```
CathPag, line 10: We really want to make sure the glossary is consistent in style, so instead of subheadings, I added them all in a paragraph. I also removed the details after errors as that was too much detail for this section

content/en/golden_signals.md
```diff
@@ -0,0 +1,24 @@
+---
+title: Golden Signals
+status: Feedback Appreciated
+category: concept
+---
+
+## What it is
+[Golden signals](https://sre.google/sre-book/monitoring-distributed-systems/#xref_monitoring_golden-signals) are metrics that help us understand how [services](https://glossary.cncf.io/service/) are experienced by users. Today, it is the gold standard for monitoring application performance and focuses on four specific metrics: Latency, traffic, errors, and saturation. The first three metrics relate to requests. A request is a message sent from one service to another. Services do that to collaborate and work as a [distributed app](https://glossary.cncf.io/distributed_apps/).
```
CathPag, line 8: Added links and an explanation of what requests are. If people are not familiar with that, the next section is hard to understand.

content/en/golden_signals.md
```diff
@@ -0,0 +1,24 @@
...
+
+*Latency* is the time it takes to service a request. *Traffic* measures the request frequency placed on the application. *Errors* measure the request failure rate. And *Saturation* measures the capacity of app resources, allowing users to address bottlenecks before they become a problem.
+ 
+## Problem it addresses
+Without metrics, engineers have nothing to deduce why applications are behaving a certain way. While earlier monitoring tools were designed to monitor machines (not applications/services), modern monitoring systems gather a huge variety of parameters. They collect so many metrics that it's easy to lose focus on what is relevant for the user. Which of all these metrics should you focus on to ensure your customers have a great user experience?
```
CathPag, line 13: You started introducing golden signal here (last two sentences). I move that to the next section.

content/en/golden_signals.md
```diff
@@ -0,0 +1,24 @@
...
+
+## How it helps
+Golden signals help engineers focus on metrics that matter for the user experience. They help steer clear from getting stuck in the mud by putting performance data into different buckets and providing the visibility needed to pinpoint problems that might affect the user.
+
+*Latency* tells us how long it takes the application to handle transactions. An increase in latency is likely a sign of a performance issue.
+
+Measuring website traffic allows a clear understanding of what is working or what isn't and develops a more potent content. It helps in better utilization of traffic.
```
CathPag, line 20: Don't think this is the right example. It's probably more about knowing when to app should scale and balance the load? But I'd defer to @iamNoah1.

CathPag (commented):
This is great, @remote007! I changed the format in the first section as we want to ensure all entries are consistent.

I also edited the three sections and added a little more context here and there so it reads better. But I believe the content is still all yours and in there (unless I noted otherwise in the comment). Let me know what you think.

@iamNoah1, could you review this PR as well?

content/en/golden_signals.md
iamNoah1, line 13:
```suggestion
Finding out why an application behaves a certain way, especially when it doesn't behave like expected, can be a hard task for engineers. While earlier monitoring tools were more focused on monitoring machines (not applications/services), modern monitoring systems gather a huge variety of metrics. It's easy to lose focus on what is relevant for the user experience. Which of all these metrics should you focus on to ensure your customers have a great user experience?
```

Rephrased the first sentence because the statement was not true. Metrics are not the only way to find out why an application behaves a certain way. There a certain other ways to achieve that. For example logging, IDE assisted live debugging even code analysis.

content/en/golden_signals.md
iamNoah1: I agree. If the traffic increases, we should scale up and the other way around. I am still not sure if we really need to explain what those metrics are.

iamNoah1 (commented):
In general, I have the impression, that the content is more complicated than needed. The baseline is that golden signals help to filter the most important metrics from all metrics available. If I have written the content, I would not have explained every metric that counts as golden signal. I would have rather thought about if we need a separate definition for them.

content/en/golden_signals.md
CathPag: Ok, great! In that case, let's end with "distributed app."  Rest looks good to you, @iamNoah1?

CathPag:
@remote007, just following up on this.

remote007:
@CathPag I thought you would be adding some suggestions here 😅.
Like opening new issues separately for four metrics and linking them here.

CathPag:
Nope, we are trying to bring this PR to a mergeable state :) If you want to continue working on this, please weave in the feedback you've received. There are a lot of unresolved comments. If something isn't clear, please ask. PRs do require a few backs and forth, that's absolutely normal.

remote007:
> In general, I have the impression, that the content is more complicated than needed. The baseline is that golden signals help to filter the most important metrics from all metrics available. If I have written the content, I would not have explained every metric that counts as golden signal. I would have rather thought about if we need a separate definition for them.

@CathPag @iamNoah1 how shall we apply this suggestion ? Open new issues for the four components and make this definition more digestible by talking only about Golden Signal .

iamNoah1:
@remote007 I would say make this definition more "digestible" like you said :). I am not sure if we need additional definition for the other terms. For now, let's focus on this PR :)

CathPag:
> > In general, I have the impression, that the content is more complicated than needed. The baseline is that golden signals help to filter the most important metrics from all metrics available. If I have written the content, I would not have explained every metric that counts as golden signal. I would have rather thought about if we need a separate definition for them.
> 
> @CathPag @iamNoah1 how shall we apply this suggestion ? Open new issues for the four components and make this definition more digestible by talking only about Golden Signal .

For now, let's just get this term ready. Let's not mention these specific metrics and keep it more general.

Issue #386, opened 2022-02-04
[Updated]: Automation


CathPag:
### Details (optional)

Here's an initial definition for [automation](https://github.com/cncf/glossary/blob/main/content/en/automation.md) that needs a few tweaks before going live. 

To claim this issue, please ask for it below AND in the #glossary channel of the CNCF Slack (to ensure we don't miss it). Before getting started, please refer to the [style guide](https://glossary.cncf.io/style-guide/).

CathPag:
There you go, @a-muk.

a-muk:
@CathPag Can you please give me some more details on this issue? I went through the links that you have provided.

JasonMorgan:
OBE

PR #418, opened 2022-02-11
made some changes in the doc- 'automation.md'

a-muk:
I made some changes in the file automation.md . These are very small changes. Please let me know if it is okay.

### DIFF
--- a/content/en/automation.md
+++ b/content/en/automation.md
@@ -6,16 +6,15 @@ category: Concept
 
 ## What it is
 
-Automation is a term for technology applications where human input is minimized.
-When any process is set-up in a way which requires less or zero human intervention is known as automation.
-For example Chatbot - You must have visited a page which as soon as you open responds with a "hi" message thus addressing your presence, that is automation. Like this automation is being used in variety of industries like medical, automobile, packaging etc.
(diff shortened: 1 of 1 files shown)


JasonMorgan:
Can we link the underlying issue this is addressing? If there is no corresponding issue please create one. Thanks!

a-muk:
@JasonMorgan https://github.com/cncf/glossary/issues/386

JasonMorgan:
Hey @a-muk after looking at a standard definition for automation, ex: https://www.merriam-webster.com/dictionary/automation, it seems like ti covers the term sufficiently well for our purposes. Can you help me understand why/if we should define it separately here?

Regardless, great attempt! Looking forward to seeing more from you as we go forward.

a-muk:
@JasonMorgan I think that the term can be defined in the beginning so that the reader can relate to the whole concept while also seeing the example given in the next line. Is that okay?

JasonMorgan:
@a-muk I think you can write a great definition of the term. My question is more should we publish a CNCF definition when the word itself is well understood outside of our context.

a-muk:
Actually it was an issue mentioned in the glossary repository. The file was already created by someone, I just made some changes in the overall explanation of the term.

PR #471, opened 2022-02-24
Update ko/virtual_machine.md (testing auto-labeling workflow)

seokho-son:
Fix #462

PR #473, opened 2022-02-24
Modify a sentence in cloud native definition clearly

seokho-son:
This PR modifies a sentence in cloud native definition clearly.

I think people who search a definition for cloud native tech may not understand what is ‘promise of the cloud’.

So, I'd like to remove ‘promise of the cloud’ and refine the sentence.

### DIFF
--- a/content/en/cloud_native_tech.md
+++ b/content/en/cloud_native_tech.md
@@ -6,7 +6,7 @@ category: Concept
 
 ## What it is
 
-Cloud native technologies, also referred to as the cloud native stack, are the technologies used to build [cloud native applications](/cloud_native_apps/). These technologiese enable organizations to build and run scalable applications in modern and dynamic environments such as public, private, and hybrid clouds, while leveraging [cloud computing](/cloud_computing/) benefits to their fullest. They are designed from the ground up to exploit the capabilities of cloud computing and containers, service meshes, microservices, and immutable infrastructure exemplify this approach.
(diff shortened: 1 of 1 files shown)


seokho-son:
I think the current setting for Labeler works for `main` branch.

jihoon-seo:
For now (with `.github/workflows/labeler.yml` like this:),
```yaml
on:
- pull_request_target
```

the Labeler workflow will be executed for all possible activity types, [which are](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target):
- assigned
- unassigned
- labeled
- unlabeled
- opened
- edited
- closed
- reopened
- synchronize
- converted_to_draft
- ready_for_review
- locked
- unlocked
- review_requested
- review_request_removed
- auto_merge_enabled
- auto_merge_disabled

And you can see that the workflow has been executed 2 times by now.
https://github.com/cncf/glossary/actions/workflows/labeler.yml
image

---

I cannot guess why the workflow was not executed when you opened the PR #471 🤔
Maybe.. if not specified, then the workflow runs only for `main` branch? 🤔

If this guess is right, then we need this option too:
```YAML
    branches:    
      - '**'
```

seokho-son:
@jihoon-seo I think we need both 'opened' and 'synchro'. 
After we push or update commits in a PR, labels can be updated too according to the files that the commits include.

> Maybe.. if not specified, then the workflow runs only for main branch? 🤔

I think so :)

CathPag (approved):
LGTM

iamNoah1 (approved):
lgtm

Issue #442, opened 2022-02-15
Add some installation hints for Hugo


iamNoah1:
This may be a small one but on the readme, setting up a local instance assumes that hugo is installed. I think it would be nice to add some info about the prerequisite of hugo that has to be installed at least. Best case IMHO there should also be some info about how to install hugo.

edsoncelio:
Great point! maybe we can have something like we have in Kubernetes docs: 
https://github.com/kubernetes/website#using-this-repository

Anosh21:
Hey I would like to prepare the installation guide for Hugo.

seokho-son:
@iamNoah1 are you interested in adding installation hints for Hugo ? or you hope to assign this task to another contributor ? :)

iamNoah1:
There we have a volunteer :) 👍

seokho-son:
@Anosh21 
Please go ahead. :)

PR #476, opened 2022-02-24
Added installation hints for Hugo (issue #442)

Anosh21:
- Added a hugo installation hint as a prerequisite.
- Added a sample link for setting up a server site.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,15 @@ All code contributions are under the Apache 2.0 license. Documentation is distri
 ## Setting up a local instance
 
 To improve the Cloud Native Glossary site itself, install a local copy with these instructions:
+## Prerequisite - Hugo Installation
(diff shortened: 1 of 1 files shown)


Anosh21:
@seokho-son

iamNoah1 (changes requested):
Thanks for the contribution @Anosh21. I think the installation hints are not accurate enough as they have additional tools as prerequisites (brew, port) and make assumptions about the underlying OS. I would rather link to an official installation page. We can leave the hints but only when we say something like "run `brew install hugo` on a mac assuming you have brew installed", at least this is my opinion. Also I would place the installation hints under the "Setting up a local instance" heading instead of introducing another heading.

iamNoah1:
fair for me, thanks @cjyabraham

PR #492, opened 2022-02-28
add Adriana to german approver team

iamNoah1:


CathPag (approved):
LGTM

jihoon-seo (approved):

iamNoah1:
You are right @seokho-son. In this case I think we can skip the whole procedure, what do you think @CathPag?. Catherine joined David and me on the german localization only because we did not have a third person until Adriana joined. She is one of the "initial" approvers and we had already a kick off meeting with her.

CathPag:
Now that we have three German approvers (including Adriana), I can be removed. I was always just a placeholder :)

iamNoah1:
> Now that we have three German approvers (including Adriana), I can be removed. I was always just a placeholder :)

done :)

PR #494, opened 2022-02-28
Add content/en/database as a service

Asmit2952:
Added a new term: Database as a Service (DBaaS)
Fixes: #340

PR #518, opened 2022-03-02
Added words related to database as a service

Asmit2952:
Added DBaas and DBA as spellchecker was failing

### DIFF
--- a/wordlist.txt
+++ b/wordlist.txt
@@ -190,3 +190,5 @@ wordlists
 www
 yaml
 yml
+DBaaS
+DBA


jihoon-seo:
How about adding another commit to the PR #494, rather than opening a new PR (I mean, this PR) ?
If you do so, you will not need to rebase the PR #494 for passing the spellcheck CI ! 😊

seokho-son:
I agree with @jihoon-seo :)

It is because we are not sure these words are required unless we don't know context. (words can be removed during review process)

Let me close this PR. @Asmit2952 please use the https://github.com/cncf/glossary/pull/494

Asmit2952:
> I agree with @jihoon-seo :)
> 
> It is because we are not sure these words are required unless we don't know context. (words can be removed during review process)
> 
> Let me close this PR. @Asmit2952 please use the #494

Sure, I'll close this PR

PR #522, opened 2022-03-03
Update automation.md

muhammedanaskhan:
Here's my first contribution - Made a term bit concise


CathPag:
I'm sorry @muhammedanaskhan, we decided not to include the term automation because a "common" automation definition will be enough to understand what it means in a technical context. Looks like we forgot to delete the file. I just submitted a PR to do so and will close this PR.

PR #523, opened 2022-03-03
Update virtualization.md

muhammedanaskhan:
corrected a spelling


CathPag:
Thanks, @muhammedanaskhan, but compute resource is actually the correct term. A computer is an actual machine. You're not the first one stumbling over this, though.

PR #525, opened 2022-03-03
Update cloud_computing.md

abs007:
Added an example for the definition for cloud computing.

### DIFF
--- a/content/en/cloud_computing.md
+++ b/content/en/cloud_computing.md
@@ -5,7 +5,7 @@ category: concept
 ---
 
 ## What it is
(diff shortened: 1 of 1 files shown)


CathPag:
Thanks, @abs007. I don't think there is a great daily life example for cloud computing. Google Drive is a [SaaS](https://glossary.cncf.io/software_as_a_service/) offering that is based on cloud computing. It's a full-blown app whereas cloud computing refers to compute resources a software engineer would use to create something like Google Drive. @iamNoah1, please correct me if I'm wrong.

iamNoah1 (commented):
> Thanks, @abs007. I don't think there is a great daily life example for cloud computing. Google Drive is a [SaaS](https://glossary.cncf.io/software_as_a_service/) offering that is based on cloud computing. It's a full-blown app whereas cloud computing refers to compute resources a software engineer would use to create something like Google Drive. @iamNoah1, please correct me if I'm wrong.

I agree. Cloud Computing is a concept and I am sorry to say, but I don't see any additional value in the content you provided @abs007 :/ as it not correct and misleading.

CathPag:
Thanks, @iamNoah1.
@abs007, I'll go ahead and close this PR. Thanks anyways.

PR #533, opened 2022-03-04
Delete automation

CathPag:
We decided not to include this term because a "common" automation definition will be enough to understand what it means in a technical context


seokho-son (approved):
LGTM

iamNoah1 (approved):

PR #543, opened 2022-03-05
Add "claiming an issue" screenshot

CathPag:


jihoon-seo (approved):
This images would be helpful 😊

Optional: I suggest to replace spaces in filename with other character, 
e.g. `claiming-an-issue.png` 😊

PR #552, opened 2022-03-08
Add term: Debugging

fsbaraglia:


content/en/debugging.md
```diff
@@ -0,0 +1,14 @@
...
+
+## How it helps
+Debugging techniques provide developers with different ways of finding bugs using special debugging tools. The usage of this tool depends on the stages of development. Analysis of logs, traces, and metrics is used for debugging directly in production without stopping the program's execution. During the programming phase, developers can use interactive debugging now on the source code. By running the program and connecting an external debugger tool, developers can control the flow of the program and instruct it to pause the execution at specific points where they suspect the error may happen; they usually get this information by reading the logs. They can understand the reason for the failure, correct the source code and create a bug fix or patch.
```
CathPag, line 14:
```suggestion
Debugging tools use various debugging techniques to help developers identify bugs. How these tools are used depends on the development stage. Analysis of logs, traces, and metrics, for instance, are used for debugging directly in production without requiring program execution to stop. Developers use interactive debugging on the source code during the programming phase. By running the program and connecting an external debugger tool, developers control the flow of the program and instruct it to pause the execution at specific points where they suspect the error may happen; they usually get this information by reading the logs. Once they have identified the source of the failure, they can correct the code and create a bug fix or patch.
```

CathPag: Just edits

content/en/debugging.md
```diff
@@ -0,0 +1,14 @@
+---
+title: Debugging
+status: Feedback Appreciated
```
CathPag, line 3: I changed it to feedback appreciated as it still needs to be reviewed by another approver

CathPag (commented):
Thanks, @fsbaraglia!  Looks good to me. Submitted a few editorial changes and have one question for the 2nd section.

content/en/debugging.md
```diff
@@ -0,0 +1,14 @@
...
+title: Debugging
+status: Completed
+category: concept
+---
+
+## What it is
+Debugging is the process or activity of finding and resolving bugs (or errors) from computer programs, software, or systems to get the desired output. A bug is a defect or a problem leading to incorrect or unexpected output.
+
+## Problem it addresses
+Developers will build the source code, create a binary file including external libraries to run it into production. They will execute it and search for errors that affect the program, which will likely not function as desired (undefined behavior). To identify the source of the problem, developers have to review the code line-by-line. Catching errors is quite hard without context, i.e., variables, data, location in memory, etc.
```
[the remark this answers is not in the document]
fsbaraglia: yes, you got perfectly the point :D I was missing some words, this is the reason that I submitted to get some ideas from you :D thank you very much

content/en/debugging.md
```diff
@@ -0,0 +1,14 @@
...
+title: Debugging
+status: Feedback Appreciated
+category: concept
+---
+
+## What it is
+Debugging is the process or activity of finding and resolving bugs (or errors) from computer programs, software, or systems to get the desired output. A bug is a defect or a problem leading to incorrect or unexpected output.
+
+## Problem it addresses
+Developers write source code and create a binary file with external libraries to run it in production. If a bug is introduced during that process, the code will likely not function as desired (undefined behavior) when executed. To identify the source of the problem, developers have to review the code line-by-line. Catching errors is quite hard without context, i.e., variables, data, location in memory, etc.
```
CathPag, line 11: Just realized one more thing, sorry. Here we determined that the problem is that we don't have any context to debug an app manually.

Presumably, debugging tools solve that by providing that needed context but we don't mention it below...or at least it's not clear how that relates to things like variables, data, location in memory, etc. 

How it helps should directly address the problem. It's probably already there but implied. We should spell it out.

CathPag (commented):
Sorry, one more comment. Just noticed when reading the entire entry.

content/en/debugging.md
```diff
@@ -0,0 +1,14 @@
+---
+title: Debugging
+status: Feedback Appreciated
+category: concept
+---
+
+## What it is
+Debugging is the process or activity of finding and resolving bugs (or errors) from computer programs, software, or systems to get the desired output. A bug is a defect or a problem leading to incorrect or unexpected output.
```
iamNoah1, line 8: I would replace output with result.

content/en/debugging.md
iamNoah1, line 11: Maybe I am too picky here, but while devs write source code, they don't necessarily create binary files. Sure, apps. run a binary at a very low level, but I think this fact does not add any value to explaining the term. Same goes  for "with external libs". 

Also it is not only about running in production. Debugging is relevant for apps running on any environment, right? 

"To identify the source of the problem, developers have to review the code line-by-line" That is not the whole truth. While it might be true, it reads as if this is the only possibility devs have. We can go through logs as well in order to find what the app is doing, without having to lock at code at all.

iamNoah1 (commented):
While reading the term definition I had the impression that we are mixing up Debugging itself as a process and IDE supported Debugging. Especially the how it helps section focuses on IDE based or to be more general tool based debugging. The other sections read more like the general debugging concept or process. My suggestion is to step back and think about the scope about this term. Is it really about debugging as a process/concept? Or is it about Tool based Debugging?

fsbaraglia:
@iamNoah1  ok, I will have a look on how to simplify. On the other hand, we should not over complicate. 

What is debugging at the end? 
A category of tools and multiple techniques to find bugs. 
Nothing more and nothing less.
Or am I wrong @CathPag  ?

iamNoah1:
@fsbaraglia sure I am not aiming for complicating things as well :) But for correctness. Consider the following two sentences.

"A category of tools and multiple techniques to find bugs."

"Multiple techniques to find bugs, which can be tool based/assisted"

My opinion is that Debugging per se is not tools, and can be done without additional tooling at all. 

But lets see what other folks think. @JasonMorgan @seokho-son @jihoon-seo @CathPag

CathPag:
> While reading the term definition I had the impression that we are mixing up Debugging itself as a process and IDE supported Debugging. Especially the how it helps section focuses on IDE based or to be more general tool based debugging. The other sections read more like the general debugging concept or process. My suggestion is to step back and think about the scope of this term. Is it really about debugging as a process/concept? Or is it about Tool based Debugging?

I think it should be about the concept itself

CathPag:
> @fsbaraglia sure I am not aiming for complicating things as well :) But for correctness. Consider the following two sentences.
> 
> "A category of tools and multiple techniques to find bugs."
> 
> "Multiple techniques to find bugs, which can be tool based/assisted"
> 
> My opinion is that Debugging per se is not tools, and can be done without additional tooling at all.
> 
> But lets see what other folks think. @JasonMorgan @seokho-son @jihoon-seo @CathPag

I'll let the technical approvers chime in. My knowledge ton that subject is very limited.

fsbaraglia:
Ok, I got you point @iamNoah1
Less tool more on concept right?

if you can please help us and make some suggestions on what to change , really appreciated

content/en/debugging.md
iamNoah1, line 11:
```suggestion
Software development is a complex activity which makes it nearly impossible to write code without introducing bugs. Those bugs lead to code that will likely not function as desired (undefined behavior) when executed. Depending on how critical an application is, bugs can have a huge negative impact, financially or even for human lives. Usually application code has to go through different stages or environments where it gets tested. The more critical an application is, the more accurate the testing has to be. 
```

content/en/debugging.md
```diff
@@ -0,0 +1,14 @@
...
+Developers write source code and create a binary file with external libraries to run it in production. If a bug is introduced during that process, the code will likely not function as desired (undefined behavior) when executed. To identify the source of the problem, developers have to review the code line-by-line. Catching errors is quite hard without context, i.e., variables, data, location in memory, etc.
+
+## How it helps
+Debugging tools use various debugging techniques to help developers identify bugs. How these tools are used depends on the development stage. Analysis of logs, traces, and metrics, for instance, are used for debugging directly in production without requiring program execution to stop. Developers use interactive debugging on the source code during the programming phase. By running the program and connecting an external debugger tool, developers control the flow of the program and instruct it to pause the execution at specific points where they suspect the error may happen; they usually get this information by reading the logs. Once they have identified the source of the failure, they can correct the code and create a bug fix or patch.
```
iamNoah1, line 14:
```suggestion
When bugs appear anywhere, engineers will have to debug (which means finding and fixing) in order to decrease undesired behavior for production systems. Debugging is no easy task as engineers have to track down the source of the undesired behavior.  It not only requires knowledge about the code itself but also knowledge about the execution context at runtime. This is were different debugging techniques and tools come in handy. Analysis of logs, traces, and metrics, for instance, are used for debugging directly in production. In contrast to that, developers can use interactive the source code during the programming phase. Interactive debugging lets developers step through the code at runtime while they can analyse the related execution context. Once they have identified the source of the failure, they can correct the code and create a bug fix or patch.
```

iamNoah1 (commented):
@fsbaraglia thanks for the patience. I had to think this one through and made some suggestions. I am keen to hear your opinion :=

content/en/debugging.md
fsbaraglia: I like it, good one @iamNoah1

fsbaraglia:
I like it , I am fine with the new version @iamNoah1  , thank you for your support

iamNoah1 (approved):
@fsbaraglia thanks for your patience and contribution :)

seokho-son (approved):
Thanks @fsbaraglia !
LGTM.

I think this PR is ready to be merged.
To make this document live, we need to change `status: Feedback Appreciated` to `status: Completed`. 
I guess @CathPag can make a final decision ;)

CathPag (commented):
I think we need to remove two words from one sentence. Please see my comment.

content/en/debugging.md
```diff
@@ -0,0 +1,14 @@
...
+
+## How it helps
+When bugs appear, engineers have to debug (e.g., finding and fixing) the app to decrease undesired behavior for production systems. Debugging is no easy task as engineers have to track down the source of the undesired behavior. It requires knowledge about the code itself and the execution context at runtime. This is where different debugging techniques and tools come in handy. Analysis of logs, traces, and metrics, for instance, are used for debugging directly in production. In contrast, developers can use interactive debugging to step through the code at runtime while analyzing the related execution context. Once they have identified the source of the failure, they correct the code and create a bug fix or patch.
```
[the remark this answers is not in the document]
iamNoah1: could be my fault, as I made a little edit because some suggestions were merged resulting in a strange sentence. I 
Agree with Catherine's suggestion :)

fsbaraglia:
> @fsbaraglia thanks for the patience. I had to think this one through and made some suggestions. I am keen to hear your opinion :=

alles klar, danke dir 💯

content/en/debugging.md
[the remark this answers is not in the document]
fsbaraglia: fixed :D thx for the suggestion

PR #631, opened 2022-03-16
Add content/en/idempotence.md

shambashib20:


content/en/idempotence.md
```diff
@@ -0,0 +1,21 @@
+---
+title: Idempotence
+status: Feedback Appreciated
+category: concept
```
CathPag, line 4: Looks like this should be a property in which case, there is no breakdown in "what it is," "problem it addresses," and "how it helps." See [style guide](https://glossary.cncf.io/style-guide/#three-subheadings).

CathPag (commented):
Thanks for your contribution, @shambashib20! Since this is a property, we need to change the format. See my comments below.

content/en/idempotence.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Idempotence
+status: Feedback Appreciated
+category: Property
+---
+
+In maths or computer science, idempotence describes an operation 
+that always leads to the same outcome, no matter how many times you execute it. If the parameters are the same, an idempotent operation won't affect the application it calls. As a language-agnostic property, it applies to any programming context.
```
CathPag, line 8: The "language-agnostic" comment sounds weird. If it's a property, why would it be language-specific unless it is the property of a programming language? But we aren't speaking about a language here.

CathPag (commented):
One additional comment. Would also need @iamNoah1 to review. This is not something I was familiar with.

content/en/idempotence.md
JasonMorgan: I agree with @CathPag please delete the last line and mark as completed. I believe this term is ready to be merged.

If you have a chance please link to another authoritative definition of Idempotence, or idempotent, as there are a few solid ones out there.

Thanks!

JasonMorgan (changes requested):

content/en/idempotence.md
shambashib20: That means just by deleting the last line, it will be okay? I will do it immediately.

content/en/idempotence.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Idempotence
+status: Feedback Appreciated
```
jihoon-seo, line 3: From @JasonMorgan's [comment](https://github.com/cncf/glossary/pull/631#discussion_r830002481):
> Please mark as completed.

```suggestion
status: Completed
```

jihoon-seo:
[Spellcheck CI fails](https://github.com/cncf/glossary/runs/5616736450?check_suite_focus=true#step:4:11).

@shambashib20 Adding a word `idempotence` to [`wordlist.txt`](https://github.com/cncf/glossary/blob/main/wordlist.txt) will solve the problem!

shambashib20:
Please help me updating the wordlist.txt. Whenever I am going to update it, it shows me a new patch branch and then it raises a new pr.
What to do?

shambashib20:
I guess I worked it out! Can you check my whole pr @JasonMorgan , @jihoon-seo ?

CathPag (approved):
LGTM

iamNoah1 (approved):
lgtm thanks @shambashib20

shambashib20:
Thanks a lot! It means a lot to me! Can you guys help me finding another issue on which I can work?

CathPag:
> Thanks a lot! It means a lot to me! Can you guys help me finding another issue on which I can work?

The best issue is the term you understand well and feel comfortable working on. Which one that is, depends on your knowledge :)

PR #677, opened 2022-04-01
[pt-BR] Update config.toml 

edsoncelio:
Update `config.toml `to fix the conflict with the PR to live
This will unblock the PR #675

PR #665, opened 2022-03-29
[hi] L10n of Feedback section

Shivansh-yadav13:
- Related issue: #470

PR #685, opened 2022-04-04
Update dev-it branch with main branch

sistella:
This PR will update dev-it branch with main branch (I merged main in dev-it)


config.toml
```diff
@@ -75,6 +75,13 @@ weight = 2
 #time_format_default = "02.01.2006"
 #time_format_blog = "02.01.2006"
 
+[languages.hi]
+title = "क्लाउड नेटिव शब्दावली"
+description = "CNCF क्लाउड नेटिव शब्दावली परियोजना का उद्देश्य क्लाउड नेटिव एप्लिकेशन के बारे में बात करते समय प्रयोग किए जाने वाले सामान्य शब्दों के संदर्भ के रूप में उपयोग किया जाना है।"
+languageName ="हिन्दी(Hindi)"
+contentDir = "content/hi"
+weight = 2
```
seokho-son, line 83: @sistella 
Is this block was merged automatically by Git?
It is not problematic, but all language would be better to use different value for `weight`. (contentDir = "content/hi" , contentDir = "content/it")

sistella: makes perfect sense! I switched it to 3, and moved it accordingly after hindi in the file

seokho-son (changes requested):
LGTM :)

config.toml
jihoon-seo: Please note that `pt-br` team opened a PR that uses the value `3` as well.
https://github.com/cncf/glossary/pull/677/files#diff-28043ff911f28a5cb5742f7638363546311225a63eabc365af5356c70d4deb77R76-R81

IMO, it would be okay to merge this PR as-is (weight: 3 for Italiano), but it is likely that merging this PR (#685) will require the PR #677 to rebase when merging into `main`, and vice versa.

Plus, in future, we have to adjust the weights of each language if some of them are using the same weight value.

jihoon-seo (approved):
LGTM!

jihoon-seo:
Preview: https://deploy-preview-685--cncfglossary.netlify.app/it/abstraction/

Feedback section is displayed like this:
image

@sistella (or another `it` contributor), Could you add some entries to [`it.toml` of `dev-it` branch](https://github.com/cncf/glossary/blob/dev-it/i18n/it.toml)?

Examples:
- https://github.com/cncf/glossary/blob/540e516abc4950f80463e3997d09355d6a671444/i18n/en.toml#L63-L71
- https://github.com/cncf/glossary/pull/665/files#diff-fd5507f0042e10570f36b7804477e5502fbb6f6a159db8e0a5e86f80c981be06
- https://github.com/cncf/glossary/pull/519/files#diff-bf91abcf6a5d6fc7b18204067500f96305db8352586d24a549a23cd973daa5a5

PTAL! 😊

seokho-son:
I think the comments from @jihoon-seo can be resolved from other PRs. (the first one can be adjusted when we merge dev-it into main)
So, let me merge this PR as is :)

PR #690, opened 2022-04-04
Localization for managed_services

sistella:
NOTE: Need to be merged after https://github.com/cncf/glossary/pull/685


sistella:
Better create it after merging the https://github.com/cncf/glossary/pull/685

PR #720, opened 2022-04-09
[bn] Need to add approvers for Bengali Localization

mitul3737:
We need more maintainers for Bengali Localization to keep the PRs checked .

PR #721, opened 2022-04-09
[bn] Need to add approvers for Bengali Localization

mitul3737:
We need more approvers to keep the PRs checked on regular basis


jihoon-seo:
@mitul3737 Could you integrate this PR to #720?
#720 and #721 is highly relevant so it is good to be in one single PR,
and it will make it easier to track the contributor updates, and the commit timeline neat. :)

Issue #756, opened 2022-04-16
Search doesn't work for foreign words


cjyabraham:
Currently the search is powered by lunr and [language support](https://lunrjs.com/guides/language_support.html) isn't activated. Lunr doesn't support Hindi, so it would never work for that language, even if we were to activate it for all other languages.

We should research a different search tool so that we can search across all foreign languages. 

Currently I am evaluating [Algolia](https://www.algolia.com/). It allows for [multiple indicies](https://www.algolia.com/doc/guides/managing-results/optimize-search-results/handling-natural-languages-nlp/how-to/multilingual-search/) which could also help with #755.

cjyabraham:
Algolia has created [an index for the Glossary site](https://www.algolia.com/apps/ZVS6RE9DZU/dashboard?).

cjyabraham:
We have switched to Google Custom Search, which solves this problem.

Issue #755, opened 2022-04-16
Search isn't language specific


cjyabraham:
When you search for a word it shows all pages on the whole site that have that word. This isn't great as it can list pages from different languages. It'd be ideal if the search only returns results for the particular language the user is in.

cjyabraham:
This is fixed in the new PageFind search

PR #763, opened 2022-04-18
Exclude localized terms from the search

jihoon-seo:
- Exclude localized terms from the search
- Delete unnecessary `.gitkeep` files

### DIFF
--- a/content/de/_TEMPLATE.md
+++ b/content/de/_TEMPLATE.md
@@ -2,6 +2,7 @@
 title: Vorlage für Definition
 status: Feedback erwünscht
 category: Konzept
+exclude_search: true
 ---
 
 ## Was es ist
--- a/content/it/_index.md
+++ b/content/it/_index.md
@@ -1,6 +1,7 @@
 
 ---
 title: "Glossario Cloud Native"
+exclude_search: true
 ---
 
 # Glossario Cloud Native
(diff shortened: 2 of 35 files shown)


cjyabraham (approved):
LGTM

seokho-son:
I don't prefer adding the config to all localized files. It is to minimize manage efforts for localized files.

I am not sure we can find anouther way to apply it globally. (By language or directory)
@cjyabraham @jihoon-seo do you have any idea ?

If we cannot find a way, we need to update localization guideline for this settting.

cjyabraham:
I see this as a short-term solution to stop non-English languages pages show up in the search. It's not a great solution because then the search on all other language pages doesn't work. We have a plan, however, to improve things more permanently:

First, I'd like to switch out the search engine to one which supports Hindi in addition to all other languages. See #756.

Second, I'd like to make it so that the search will only search pages that are in the current language. See #755.

I'll keep everyone updated with progress on those issues.

seokho-son:
I believe a permanent solution will be applied. :) 
I hope to bring this issue to a maintainers' meeting to discuss whether this short-term solution is required.

seokho-son:
Maintainers agreed to stay in current status until a solid solution is applied. 
Let's hold this PR until we see some progress on #756, #755.

seokho-son:
Maintainers agreed to stay in current status until a solid solution is applied. 
Let's hold this PR until we see some progress on #756, #755.

PR #795, opened 2022-04-27
[fix] Translate all category template fields

electrocucaracha:
We have recently discovered that category fields need to be translated, so this change modifies all the existing Spanish translations as well as the word list file.


krol3 (approved):
LGTM

raelga (approved):

Issue #410, opened 2022-02-08
Tracking Bengali Localization Progress of Glossary


mitul3737:
This issue tracks Bengali localization progress.

# Join the Slack

- Join the [CNCF Slack](https://cloud-native.slack.com/join/shared_invite/zt-11yzivnzq-hs12vUAYFZmnqE3r7ILz9A#/shared-invite/email)
- Join the [glossary](https://cloud-native.slack.com/messages/glossary), [ glossary localization channel](https://cloud-native.slack.com/messages/glossary-localizations) , [Bengali localization channel](https://cloud-native.slack.com/messages/glossary-localization-bengali)

# Checklist to translate a term
1.	Check that no one else has claimed the page in the comments below.
2.	Comment below the page that you would like to localize, focus on one page per PR, at a time.
2.	Comment below the page that you would like to localize, focus on one page per PR, at a time.
3.	Maintainers will add your username and Pull Request next to your selected page in the below checklist.
4.	Fork this repo, make your localizations, and submit a pull request following the name convention ``[bn]  Localize <file_name.md>``.
ex: [bn] Localize cloud_native_apps.md
5.	tag/ping a maintainer to make it easier to track progress

# Points to remember:

- Please use the word localization instead of translation.

- For more information on how to contribute, check out [guidelines](https://glossary.cncf.io/contribute/)

# For Volunteers
Volunteers can comment below on this issue with the page-name they are interested in localizing.
Please also provide a placeholder PR in your comment.
Example: ``I would like to contribute to content/bn/<xyz.md>``

# To create a PR:
To create a PR:

- Fork https://github.com/cncf/glossary/tree/dev-bn repository to your personal account

- Create a branch with name suitable for your assigned task ``e.g. Localize content/bn/style-guide/_index.md``

- Once the changes are made, create a pull request (PR) against ``cncf:dev-bn`` from ``<your-github-id>:<branch-name>``

For example:

![O_pr](https://user-images.githubusercontent.com/57193846/153064308-ac3aecc3-e513-4a74-86cb-e056427b084e.png)

- Mention this issue link in the PR description

- NOTE: Please create 1 PR for a single page from tasks below. This makes it easier to track progress.

# Priorities
We need at least 10 terms in Hindi before go live, below are some suggested terms to work:

Minimum 10 terms required:

- [ ] config.toml
  content/bn/_index.md
  content/bn/cloud_native_tech.md
  content/bn/contribute/_index.md
  content/bn/style-guide/_index.md
[...]

PR #840, opened 2022-05-06
[bn] Feedback localization for "yes" and "no" option

mitul3737:
This PR covers changes to Localize the Feedback option properly.
While pressing "yes" in the Feedback, the user will get a message in Bengali and same while pressing "no"
Tracking Issue: #410


Imtiaz1234 (approved):
lgtm

seokho-son:
Hi @mitul3737 

We should not merge this PR.

config.toml is global configuration for all l10n contents including English contents.

The message after feedback button cannot be localized yet. 

So, let me close this PR.

PR #871, opened 2022-05-09
Adopt semantic line breaks

jihoon-seo:
This is kind of a pilot PR for adopting semantic line breaks into the markdown files, which will make the review process easier.
After collecting feedbacks and eventually merging this PR, I will work on other markdown files as well. 😊


iamNoah1 (approved):
lgtm

CathPag (approved):
Wow! Thanks for doing that, @jihoon-seo.

Issue #767, opened 2022-04-20
Add tags


cjyabraham:
Activate the tagging feature in docsy. Ideally these tags will go on top of the categories.

cjyabraham:
Here are the design options from Gareth:
<img width="971" alt="glossary1" src="https://user-images.githubusercontent.com/463160/164459707-ff0aaf93-81dc-4ef4-a5ae-a715eb06363c.png">
<img width="971" alt="glossary2" src="https://user-images.githubusercontent.com/463160/164459733-f197f1dc-6ca9-489a-85e0-e6e189e09f11.png">

CathPag:
@cjyabraham, I updated the order of the tags. Fundamentals and Advanced should come first, then Technology, Property or Concept, and then the rest — thank you!! 


1. Fundamentals
2. Advanced
3. Technology
4. Property
5. Concept
6. Application
7. Infrastructure
8. Architecture
9. Networking
10. Data
11. Security
12. Platform
13. Analytics
14. Methodology

CathPag:
> Here are the design options from Gareth: <img alt="glossary1" width="971" src="https://user-images.githubusercontent.com/463160/164459707-ff0aaf93-81dc-4ef4-a5ae-a715eb06363c.png"> <img alt="glossary2" width="971" src="https://user-images.githubusercontent.com/463160/164459733-f197f1dc-6ca9-489a-85e0-e6e189e09f11.png">

I would 100% go with pink. It needs to pop so people see it.

PR #930, opened 2022-06-08
Some design and search improvements

cjyabraham:
- changes topnav to be slightly transparent to match www.cncf.io design
- styles the search input box to comply with new design
- repositions language dropdown to be not off-screen
- swaps in Google search instead of local search to allow for foreign language word searches ([example](https://deploy-preview-930--cncfglossary.netlify.app/search/?q=api))
- added tags for articles ([example](https://deploy-preview-930--cncfglossary.netlify.app/auto_scaling/))
- creates articles by tag page ([example](https://deploy-preview-930--cncfglossary.netlify.app/tags/infrastructure/))


jihoon-seo:
[Previews]
- Search result of "virtual machine": https://deploy-preview-930--cncfglossary.netlify.app/search/?q=virtual%20machine
- Search result of "가상 머신" (which is "virtual machine" in Korean): https://deploy-preview-930--cncfglossary.netlify.app/search/?q=가상%20머신

jihoon-seo:
@cjyabraham We might delete the line `exclude_search: true` from `content/hi/**.md` files,
since that field will be not effective anymore. 😊

jihoon-seo:
Action items
- to maintainers: 
  - add tags to each English terms
- to L10n teams: 
  - add tags to each localized terms
  - add `content/**/search.md` of their own

seokho-son:
@cjyabraham Thank you for the improvement !! 

image

`Edit this page` -> https://github.com/cncf/glossary/edit/main/content/en/search.md
` Report issue` -> https://github.com/cncf/glossary/issues/new?title=Search%20Results

in Search results page.

should we remove them ?

seokho-son:
Here are some questions :)

1) Can we apply a relative path to a `tag` button ?

    image
  
    Current button for a tag is linked to https://glossary.cncf.io/tags/ and not to https://deploy-preview-930--cncfglossary.netlify.app/tags . 

2) How readers can find the list for existing tags easily (list in https://deploy-preview-930--cncfglossary.netlify.app/tags/) ?
    image

3) tags: ["fundamentals", "platform", ""] is possible to localize ? (ex: tags: ["기초", "플랫폼", ""] )

cjyabraham:
@seokho-son for (3), yes, when people add tags to non-English pages they will add them in their own language.

cjyabraham:
(1) should be fixed.

For (2), I've added an "All tags" link at the bottom of [a tag page](https://deploy-preview-930--cncfglossary.netlify.app/tags/fundamentals/).

cjyabraham:
There's a new label that needs translating `ui_see_all` for all languages other than English.

CathPag:
One more thing, the initial categories (technologies, concept, property) also need to be converted into tags (they go right after fundamentals/advanced). [Here's the list ](https://github.com/cncf/glossary/issues/767#issuecomment-1107477162)of the tag order. That means every term has to have at least two tags. 

That also frees up space above for all the tags.

cjyabraham:
I'm a bit confused. Tags are normally independent from each other. Categories, on the other hand, are mutually exclusive. I think that the initial categories (technologies, concept, property) should remain in their own taxonomy and remain mutually exclusive, while tags can behave just as normal tags. I think it's better to go that route as it complies with how people would expect tags and categories to behave. What do you think?

CathPag:
> I'm a bit confused. Tags are normally independent from each other. Categories, on the other hand, are mutually exclusive. I think that the initial categories (technologies, concept, property) should remain in their own taxonomy and remain mutually exclusive, while tags can behave just as normal tags. I think it's better to go that route as it complies with how people would expect tags and categories to behave. What do you think?

I don't think they need to be mutually exclusive. Some will fall under two or more categories — it's not as strictly divided. The best example is fundamentals/advanced. Those will all be in two categories at least. Here we want to give people the ability to filter for basic terms to make sure they understand them...or let them know that a term is advanced and may not be easily understood without previous technical knowledge. Maybe we should color code them: fundamentals and advance one color (or pink shade), tech/concept/property in a different one, and the rest in a third color or shade. Each of these tags tells you a little bit about the term prior to reading the definition and that's valuable, don't you think? 

Happy to merge as is and update in the next iteration.

seokho-son:
@cjyabraham 
How about we merge this PR just after an update to fix the conflicts ?

As @CathPag suggested, we can discuss issue regarding Tags and Categories later, and update in the next iteration.

iamNoah1:
Agreed with @CathPag and @seokho-son about merging this and clarifying tags vs categories afterwards. 

My two cents on tags vs categories is that we need something that is not mutually exclusive and I don't really care if we call if tag or category. Kubernetes  glossary calls it tags, so I assume we should that way too. https://kubernetes.io/de/docs/reference/glossary/?fundamental=true&extension=true&operation=true

cjyabraham:
Ok, cool. I am on vacation next week but will try to get this finished when I get back.

cjyabraham:
This is ready to merge now

jihoon-seo (approved):
LGTM! What a great work!

For who merge this PR: it would be great to merge this with the merge method "Sqaush and merge". 😊

PR #943, opened 2022-06-15
[zh] fix terminologie and format

lunarwhite:


hanyuancheung:
`dev-zh` branch is the base one for Chinese language, pls commit to it instead of the main branch.

PR #977, opened 2022-06-20
Term/distributed apps

ugho16:


sistella (changes requested):
please branch off from dev-it branch

PR #992, opened 2022-06-25
Added Virtual Private Cloud

akshatcoder-hash:


iamNoah1 (changes requested):
Thanks for your contribution @akshatcoder-hash. 

Some feedback: 
* Please use semantic line breaks
* Please don't use bullet points, instead write complete sentences

PR #1004, opened 2022-06-27
expands and refines o11y definition

mhausenblas:
With this PR I'm trying to make the definition of the term `observability` a bit more tangible and refine the scope and purpose of it. 

Signed-off-by: Michael Hausenblas <hausenbl@amazon.com>


CathPag:
Thanks for the update, @mhausenblas — it looks great! Only one thing, your definition focuses on an observability tool's **_capability_** to observe an observable app (that's a mouth full!). This definition focuses on the app's **_property_**. Do you mind adjusting it so it changes the perspective back to the app and its property?

mhausenblas:
Hey @CathPag thanks a lot for the feedback! I have no idea why I didn't get the notification about your response, hence the delay.

> Do you mind adjusting it so it changes the perspective back to the app and its property?

Hmm I'd love to do that but I'm not 100% sure I understand the ask. Adjusting what? :)

Maybe my phrasing was unclear but I'm not arguing for the tool capability but the systems (system under observation vs the tooling used to do telemetry and destinations to consume said signals). Does this make sense?

CathPag:
Hi @mhausenblas, sorry, let me try to explain what I mean.  Observability is a _property_ of an application that allows you to determine the internal state of a system through external outputs. To be observable, the app must generate data observability tools can analyze. 

Your definition seems to focus on observability **_tools_** and their **_capability_** of making sense of that data (no the property of the app that generates this data):

> Observability is the capability to continuously generate and discover actionable insights, based on signals from the system under observation, that is, derive the system's state from its external output with the goal to influence the system.

You do say "generate" which is what the app does but then move right back to the tool "discover actionable insights...derive the system's stage from its external output." It'd be great if this definition could focus on the property and then, in the how it helps section, talk about how observability tools help us understand what's going on.

In short, I was asking if you could change the focus from the observability tool (a technology with certain capabilities) to the app's property (not a specific technology). Does that make sense? 

We could add an "observability tools" definition that would build on this definition to focus on technologies and their capabilities specifically.

mhausenblas:
Thanks for your a explanation @CathPag and in fact it's the opposite I'm arguing for :)

As in: I don't agree that observability is "just" a property of an app, it's the capability  of the overall organizational system which includes tools AND humans. Happy to meet up and discuss this, maybe things get lost here in the comments?

CathPag:
Ok, let me ping some of the other maintainers so they can join the discussion. I'm the only one who hasn't a technical background and was going with what we had previously agreed upon. 

We do have a public meeting on Thursday at 9 am EST (details are on the [CNCF calendar](https://www.cncf.io/calendar/)). Since you offered to meet up, we could put that on the very top of the agenda (if a discussion via comments doesn't get us where we need to be). 

Excited to see discussions happening here, btw. This is exactly what we need!

caniszczyk:
I'm fine with @mhausenblas updated definition

CathPag (commented):
Ok, Chris confirmed so let's move forward :) We'll need a few changes, though:

1. Update category to concept 
2. Since we aren't framing this as a property, we need to update the format and include missing sections >> we need something for "problem it addresses"
3. We do per sentence line breaks to facilitate edits: I updated everything as suggested changes. It looks like a bunch of text changes but it's mostly just connecting sentences

CathPag:
Awesome! Yes, we need to update the header to: `category: concept` We have three broad categories: Technology, concept, and property. Since property is a characteristic, we don't have a "problem it addresses" and "how it helps" section.

[Chaos Engineering](https://glossary.cncf.io/chaos-engineering/) is a good example:

- **What it is**: clear statement of what we are talking about (for some, this might be all they need to know. Let's call it minimal viable definition...well, minimal viable is a little too much but you get what I mean :) 
- **Problem it addresses**: doesn't even mention CE. It describes a world without it and the pain points that exist (that's why it doesn't make sense for properties)
- **How it helps**: explains how this concept or technology addresses that problem head-on

I hope this helps.

CathPag (approved):
Love it! To the point and easy to understand — thank you so much for your contribution, @mhausenblas!

iamNoah1 (approved):
I agree with Catherine, nice content @mhausenblas. Thanks for the contribution :) 

lgtm

PR #1025, opened 2022-07-01
[zh]merge main into dev-zh 

Petrie:
All docs added `tags` . We need merge `main` into `dev-zh` to avoid conflicts in the future


Petrie:
I have already signed-off the `git merge` commit.  why still can't pass the DCO.

PR #922, opened 2022-06-02
Rename markdown filenames

jihoon-seo:
This PR
- closes #799.
- renames markdown filenames, replacing all the underscores (`_`) with hyphen (`-`).
  - "no capitalization, no spaces, no parentheses, and use hyphens to separate words"

---

[Commits]
- Insert empty lines before and after the header lines (for all languages)
- Add semantic line breaks (for English files only)
- Rename markdown filenames (for all languages)
- Update relevant links (for all languages)
- ...

PR #1037, opened 2022-07-04
[ko] Add tags to each localized term

jihoon-seo:
- Add tags to each localized term (from https://github.com/cncf/glossary/pull/930#issuecomment-1149681643)
- Add semantic line breaks (from #922)


seokho-son:
PTAL @yunkon-kim

yunkon-kim (approved):
LGTM !!

PR #1042, opened 2022-07-05
[ptbr] Add search.md to portuguese content

edsoncelio:
Adding the `search.md` file based on this https://github.com/cncf/glossary/pull/930#issuecomment-1149681643

### DIFF
--- a/content/pt-br/search.md
+++ b/content/pt-br/search.md
@@ -1,0 +1,4 @@
+---
+title: Resultados da Busca
+layout: search
+---


jihoon-seo:
In preview, if I search for 'virtual machine' at Glossary `pt-br` home page (https://deploy-preview-1042--cncfglossary.netlify.app/pt-br/),
it goes to https://glossary.cncf.io/search/?q=virtual%20machine .

What would be great:
it goes to https://deploy-preview-1042--cncfglossary.netlify.app/pt-br/search/?q=virtual%20machine (staying in `pt-br` lang),
or at least, https://glossary.cncf.io/pt-br/search/?q=virtual%20machine .

@cjyabraham could you please take a look whether we can apply some modification for this?

cjyabraham:
@jihoon-seo Currently there is no way to limit search results to just a particular language so I don't really see the point of nesting the search results within different language subfolders. That might give the impression that the results are language-specific when they are not. If it's a matter of localizing the words "Search Results" in the title of the page we may be able to do that instead?

I've just created a PR #1050 which is not what you're asking for, however, at least keeps search on the same dev instance rather than redirecting to the live site on search form submit. I admit it's not what you are requesting but nice to fix that problem.

jihoon-seo:
@cjyabraham I got your point. Thanks for your enhancement suggestion.

I suggested [this](https://github.com/cncf/glossary/pull/1042#issuecomment-1175776603) in UX perspective:
let's suppose there exists a non-English speaker.
In the middle of his navigating Glossary website in his language (for example, `ko`), 
if he performs a search for some term,
then the website fallbacks to English mode (sidebar, topnav, content), which might degrade the user experience.

---

Some other reference:
I found that k8s.io's search function does not sustain the current language, but fallbacks to English mode.
- https://kubernetes.io/ko/docs/home/

cjyabraham:
@jihoon-seo I see your point and agree what you propose would be a better experience. I've been trying to modify [the search.js file](https://github.com/google/docsy/blob/170cdd9cd31794eac802b0cec083097155feb81b/assets/js/search.js) so that line 30 reads:
`                    var searchPage = "{{ "search/" | absLangURL }}?q=" + query;`
However, the `absLangURL` function doesn't seem to be having any effect in keeping the search within the same language directory.

Additionally, the `{{ .Lang }}` var always seems to return `en` even when I'm in a `hi` file. So perhaps this whole system is still buggy.

I'm not sure where to go from here so would welcome any help with this.

PR #1044, opened 2022-07-05
[zh]translate debugging into chinese

Petrie:


hanyuancheung (approved):
lgtm, thanks for @Petrie ~!

Rocksnake (approved):

PR #1050, opened 2022-07-06
Get Search working on PR instance

cjyabraham:
Currently the search form on PR instance redirects to the live Glossary site. This PR is to get it so that the form keeps the user on current instance.

If you go to [this dev instance](https://deploy-preview-1050--cncfglossary.netlify.app/) and search for something, you'll see the search results now appear on the same dev instance rather than going to the live site.

Signed-off-by: cjyabraham <cjyabraham@gmail.com>

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -1,4 +1,4 @@
-baseURL = "https://glossary.cncf.io/"
+baseURL = "/"
 title = "Cloud Native Glossary"
 
 enableRobotsTXT = true


cjyabraham:
You tested this instance? https://deploy-preview-1050--cncfglossary.netlify.app/

CathPag:
Yep, just send you a screenrecording via Slack

CathPag (approved):
LGTM

jihoon-seo:
@cjyabraham would there be no side effect from replacing `baseURL` in `config.toml` from `"https://glossary.cncf.io/"` to `"/"`? 😊

cjyabraham:
I don't think so as that is the way we do it over on [the Contribute site](https://contribute.cncf.io/): https://github.com/cncf/tag-contributor-strategy/blob/main/website/config.toml#L1

We should definitely test it thoroughly when we deploy.

jihoon-seo (approved):
LGTM!

PR #1081, opened 2022-07-14
Glossary style tweaks

thetwopct:
Related to this [issue](https://github.com/cncf/cncf.io/issues/562) as part of brand update for CNCF:

- Aligned the style with Figma designs a bit more
- Removed some bootstrap classes from the code where we were overriding them in CSS
- Tidied up the styles and made them a bit more organised in to separate SCSS files
- Utilised more use of Bootstrap CSS options so we do not need to override them in CSS
- Copied across the CNCF custom hamburger menu button as the Font Awesome one didn't look great
- Added fade to the top and bottom of the sidebar when there is overflow (not super happy with how this looks at the top - it looks strange with the underline of the search bar IMO)
- Currently there is no way to switch languages on mobile, so I added language selector to the footer on mobile
- Experimented with changing the hamburger icon on mobile menu open/close but reverted as Bootstrap doesn't seem to handle the add/removal of collapsed class properly

Did not replace all icons this time around 1) some due to unresolved decisions around Categories/Tags, and 2) replacing Font Awesome for custom icons like Edit or Github seems overkill when Font Awesome icons are already pretty good, we could use the orgs Font Awesome Pro licence, or even [icons8](https://icons8.com/line-awesome) (which seems to be match our styling more than FA) 

Regarding custom styles overriding Docsy theme - I've changed the way this was working. Previously this had all our custom variables and CSS in `_variables_project.scss`. This file is designed for project based variables as it get loaded before other theme stylesheets. There is another file where custom CSS should go - `_styles_project.scss` - this is loaded after all other theme files, so there are less problems with specificity and less need to use !important on statements.


cjyabraham:
Thanks @thetwopct. Good to know about `_styles_project.scss`. I wasn't happy with all the `!important`s we were using.

One issue I don't know if there's a quick fix to is that the category icons are not appearing for non-English languages. [For example](https://deploy-preview-1081--cncfglossary.netlify.app/pt-br/blue-green-deployment/). This is due to the class being applied is using the translated category name which has no match in the CSS. We can break this into a separate issue if there isn't a quick fix for it.

thetwopct:
> if there isn't a quick fix for it

I can't find one from my initial searching.

1) Adding each translated category as a CSS class is not a scalable solution
2) I had an idea to always pull the EN version of the category string as the class name, but [seems Hugo cannot do that yet](https://github.com/gohugoio/hugo/issues/7844)
3) It also doesn't seem possible to switch languages in the HTML template from source to then output the EN name of the category and then switch back to the source.
4) We could also add an icon_name param to the front matter of every version of every post, but that seems overkill.

Maybe we ask for input on https://discourse.gohugo.io

jihoon-seo (approved):
Quick review:

> - Aligned the style with Figma designs a bit more
> - Tidied up the styles
> - Copied across the CNCF custom hamburger menu button as the Font Awesome one didn't look great

[As-is] (mobile)
image

[To-be]
image

---

> - Added fade to the top and bottom of the sidebar when there is overflow (not super happy with how this looks at the top - it looks strange with the underline of the search bar IMO)

[As-is] (PC)
image

[To-be]
image

---

> - Currently there is no way to switch languages on mobile, so I added language selector to the footer on mobile

[As-is] (mobile)
image

[To-be]
image

LGTM!
Thank you for your hard work @thetwopct!

CathPag:
Yes, thank you for the improvements, @thetwopct, and thank you for the screenshots, @jihoon-seo. I have to admit, I didn't always understand what James was referring to :)

CathPag (approved):
LGTM

PR #1095, opened 2022-07-19
[zh] translate `bare metal machine`

Petrie:
Signed-off-by: Petrie <lpfvip2008@gmail.com>


Jacob953 (approved):
It would be better to be segmented for readability. :)

Petrie:
> It would be better to be segmented for readability. :)

Got it !

hanyuancheung (approved):

PR #1104, opened 2022-07-22
[zh]add tags

Petrie:
`tags` generated with:
```sh 
cd content
for i in `ls en/` ; do  tags=`grep 'tags' en/$i`; sed -i "5i $tags" zh-cn/$i ;done;
```
Signed-off-by: Petrie <lpfvip2008@gmail.com>


Jacob953 (approved):
LGTM. Thx a lot. ;)

hanyuancheung (approved):

PR #1120, opened 2022-07-26
[it] Fix tls cd autoscaling

fsbaraglia:


fsbaraglia:
needed to cleanup the local branch before and add sign-off

PR #1163, opened 2022-08-04
Update wordlist.txt to add ci/cd

HenokB:
Signed-off-by: Henok Biadglign Ademtew <46082799+HenokB@users.noreply.github.com>


iamNoah1 (commented):
Hi @HenokB, I don't see why you created this PR. The wordlist gets updated as soon as the wording is included in a term definition.

PR #1171, opened 2022-08-05
Update _index.md

annalisag-spark:
I updated the Acknowledgement paragraph, in order to add the new contributors and delete inactive people.


iamNoah1 (changes requested):
@annalisag-spark seems like your PR contains a lot more commits than the changes you mentioned. I guess you based your changes on main branch instead of dev-it.

annalisag-spark:
@iamNoah1 I noticed after I wrote it. But what shall I do? I follow the GitHub path that ignited when you click on the CTA "Modifica pagina" from this page: https://glossary.cncf.io/it/

How does it work then? Is this to be avoided? Thanks for explaining and sorry for the troubles caused.

iamNoah1:
@annalisag-spark So what happens when you go that way, is that the changes you make are based on the main branch, so you have to choose the main branch again as target for your PR. However, the target branch was set to dev-it. I Changed it. Usually I guess the changes should be based on dev-it and then propagated to main at some point. Now we have to merge main do dev-it only then changes will apply for you localized PRs. I hope this was understandable ^^
Comparing_main___iamNoah1-patch-1_·_cncf_glossary

annalisag-spark:
@iamNoah1 😅 not really. If those changes are based on the main by default, but we have to target dev-it (which is the correct and agreed upon process) can we say that working directly from the website CTAs is not recommended? It would be better to work on the repo files following the correct dev-xx process, correct? (Just to understand the best way to operate). 
Thanks for your patience

iamNoah1:
@annalisag-spark exactly, localization efforts should be done on dev-it. I guess it is not possible to choose base branch when using the github UI, so the recommendation is to not use it especially when dealing with more sophisticated operations.

fsbaraglia (approved):

fsbaraglia (approved):

PR #1181, opened 2022-08-07
Some fixes

cjyabraham:
Some misc fixes:
- The sitemap was listing all terms, not just the ones that are "Completed". This is now fixed.
- There were some code formatting issues that were preventing my hugo from compiling and running locally.

### DIFF
--- a/layouts/_default/sitemap.xml
+++ b/layouts/_default/sitemap.xml
@@ -1,0 +1,24 @@
+{{ printf "<?xml version=\"1.0\" encoding=\"utf-8\" standalone=\"yes\"?>" | safeHTML }}
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9"
+  xmlns:xhtml="http://www.w3.org/1999/xhtml">
+  {{ range ( where .Data.Pages ".Params.status" "Completed" ) | union ( where .Data.Pages ".Params.status" "completed" ) }}
+    {{- if .Permalink -}}
+  <url>
+    <loc>{{ .Permalink }}</loc>{{ if not .Lastmod.IsZero }}
+    <lastmod>{{ safeHTML ( .Lastmod.Format "2006-01-02T15:04:05-07:00" ) }}</lastmod>{{ end }}{{ with .Sitemap.ChangeFreq }}
+    <changefreq>{{ . }}</changefreq>{{ end }}{{ if ge .Sitemap.Priority 0.0 }}
+    <priority>{{ .Sitemap.Priority }}</priority>{{ end }}{{ if .IsTranslated }}{{ range .Translations }}
+    <xhtml:link
+                rel="alternate"
+                hreflang="{{ .Language.Lang }}"
+                href="{{ .Permalink }}"
+                />{{ end }}
+    <xhtml:link
+                rel="alternate"
+                hreflang="{{ .Language.Lang }}"
+                href="{{ .Permalink }}"
+                />{{ end }}
+  </url>
+    {{- end -}}
+  {{ end }}
+</urlset>
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -103,8 +103,8 @@
 				Documentation Distributed under CC BY 4.0</small>
 			{{ if not .Site.Params.ui.footer_about_disable }}
 			{{ with .Site.GetPage "about" }}
-			<p class="mt-2"><a href="{{ .RelPermalink }}">{{ .Title }}</a></p>{{
-			end }}
+			<p class="mt-2"><a href="{{ .RelPermalink }}">{{ .Title }}</a></p>
+			{{ end }}
 			{{ end }}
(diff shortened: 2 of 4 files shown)


seokho-son (approved):
LGTM :)
Thanks !

layouts/_default/sitemap.xml
```diff
@@ -0,0 +1,24 @@
+{{ printf "<?xml version=\"1.0\" encoding=\"utf-8\" standalone=\"yes\"?>" | safeHTML }}
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9"
+  xmlns:xhtml="http://www.w3.org/1999/xhtml">
+  {{ range ( where .Data.Pages ".Params.status" "Completed" ) | union ( where .Data.Pages ".Params.status" "completed" ) }}
```
jihoon-seo, line 4: [`en` Sitemap]
- As-is: https://glossary.cncf.io/en/sitemap.xml
- To-be: https://deploy-preview-1181--cncfglossary.netlify.app/en/sitemap.xml

This change also removes these URLs from the sitemap, (which might impact on the Google search results):

- Some guides
  - https://glossary.cncf.io/contributor-ladder/
  - https://glossary.cncf.io/contribute/
  - https://glossary.cncf.io/style-guide/
- URLs for each tag
  - https://glossary.cncf.io/tags/
  - https://glossary.cncf.io/tags/application/
  - https://glossary.cncf.io/tags/architecture/
  - ...
- The "Categories" page
  - https://glossary.cncf.io/categories/

If needed, we might add the front matter `status: Completed` to the "Contribution Ladder", "How to Contribute", and "Style Guide" markdown files (and so on), and see if the result changes.
@cjyabraham what do you think about this? 😊

jihoon-seo (commented):
@cjyabraham PTAL!

layouts/_default/sitemap.xml
cjyabraham: Thanks for catching this. I have now changed the logic to include the above pages again. I now only omit the page if they have `status: Feedback Appreciated`.

jihoon-seo: > I now only omit the page if they have `status: Feedback Appreciated`.

Sorry for bothering you @cjyabraham, but there exists a possibility that a contributor specifies other than exact `Feedback Appreciated` in the `status:` front matter. (For now, your suggested logic seems case-sensitive.) 😅
- `Feedback appreciated`
- `feedback appreciated`
- `Feedback erwünscht` (this example is from [`content/de/_TEMPLATE.md`](https://github.com/cncf/glossary/blob/main/content/de/_TEMPLATE.md?plain=1))
  - other lang contributors might specify `Feedback Appreciated` in their langs as well, 
    and L10n PRs like these could be merged into each lang's dev branch without maintainers' knowing.
- ...

And for consistency with other parts of the site (`layouts/partials/sidebar-tree.html`),
I would prefer your original approach: 498b021.

@cjyabraham what do you think would be a good solution for this issue? 😊

cjyabraham: This is a fair point. Ok, I have reverted that last change and have also added `status: Completed` to the pages that were omitted previously. I'm not worried about the tag and category pages being listed in the sitemap. They'll be indexed by Google anyway.

jihoon-seo: LGTM!

Just for information:
the tag and category pages will **not** be listed after this PR gets merged (see [this](https://deploy-preview-1181--cncfglossary.netlify.app/en/sitemap.xml)),
but IMO either one would be okay, those being listed in the sitemap or not. 😊

jihoon-seo (approved):
LGTM!

layouts/_default/sitemap.xml
cjyabraham: My understanding is that being omitted from a sitemap doesn't mean that Google won't list a page, it's just that Google won't _discover_ that page by way of the sitemap. Google can also discover pages by following links to them from other pages. I thought that since Google already indexes the tags and category pages it will continue to do so, even though they no longer are in the sitemap. Is your understanding different?

cjyabraham: Here's [some info on that](https://developers.google.com/search/docs/advanced/sitemaps/overview#do-i-need-a-sitemap).

hanyuancheung (approved):

Rocksnake (approved):

mitul3737 (approved):
LGTM

raelga (approved):

layouts/_default/sitemap.xml
jihoon-seo: > Is your understanding different?

No, my understanding is not different from yours. I just wanted to say that I agree with you @cjyabraham. 😊

CathPag (approved):
LGTM

seokho-son (approved):
LGTM. Let me merge this PR.

PR #1286, opened 2022-09-02
Enable labeler workflow

seokho-son:
Signed-off-by: Seokho Son <shsongist@gmail.com>

### Describe your changes
- This PR is to update labeler workflow (automate labeling PRs)
  - https://github.com/actions/labeler#create-workflow
- The labeler workflow has been disabled to check a security issue within the labeler workflow. (a high level permission given to the `pull_request_target` option)
- According to some valuable discussions with @yunkon-kim , we concluded that the concern with the permission issue can be minimized. (please review and discuss if there is any other concern)

- NA

  - [signing off](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) (ex: `git commit -s`) is to affirm that commits comply [DCO](https://wiki.linuxfoundation.org/dco).

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,14 +1,12 @@
 name: "Pull Request Labeler"
 on:
-  pull_request:
-    types: [opened, synchronize]
-    paths:
-      - 'content/**'
-    branches:
-      - '**'
+- pull_request_target
 
 jobs:
   triage:
+    permissions:
+      contents: read
+      pull-requests: write
     runs-on: ubuntu-latest
     steps:
     - uses: actions/labeler@v4
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -6,6 +6,7 @@ on:
       - 'content/**.md'
     branches:
       - 'dev-**'
+      - 'main'
 
 jobs:
   triage:
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,6 +1,11 @@
 name: "Pull Request Labeler"
 on:
-- pull_request_target
+  pull_request_target:
+    types: [opened]
+    paths:
+      - 'content/**.md'
+    branches:
+      - 'dev-**'
 
 jobs:
   triage:


seokho-son:
PTAL @yunkon-kim

.github/workflows/labeler.yml
```diff
@@ -1,16 +1,14 @@
 name: "Pull Request Labeler"
 on:
-  pull_request:
-    types: [opened, synchronize]
-    paths:
-      - 'content/**'
-    branches:
-      - '**'
+- pull_request_target
```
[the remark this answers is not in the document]
yunkon-kim: It's very cool from the human labeler point of view 😍 

Currently, this workflow is triggered by every PRs, and workflows triggered via 'pull_request_target' have write permission to this upstream repository. 

Therefore, IMHO, it would be a good for me to explicitly or restrictively handle this workflow. (i.e. only for labeling PRs related to localization). Later we can consider gradually expanding the scope of this workflow later.

yunkon-kim (commented):
@seokho-son I have a simple suggestion as follows :-)

.github/workflows/labeler.yml
jihoon-seo, line 3: Amending to @yunkon-kim's [suggestion](https://github.com/cncf/glossary/pull/1286/files#r962308875):

```suggestion
  pull_request_target:
    types: [opened, synchronize]
    paths:
      - 'content/**.md'
    branches:
      - 'dev-**'
      - 'main'
```

If we add 'main' as target branch, then this will automatically label `lang/en` for the PRs that add/modify 'content/en/**.md' and target `main`!

(Ref: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-based-on-the-head-or-base-branch-of-a-pull-request-1)

---

For information:
The default value of `sync-labels` is `false`. (https://github.com/actions/labeler#inputs)
We might need to add [`action.yml`](https://github.com/actions/labeler/blob/main/action.yml) to this repo and set the value of `sync-labels` to `true`, 
to make `synchronize`-related thing work.

yunkon-kim: @jihoon-seo thanks for your updates!

I agree that `main` is a vital point in this workflow.

For a simple discussion:
I was thinking about the `synchronize` type. However, I couldn't find any cases of reassigning the label when synchronizing.

Also, when we add the `synchronize` type, this workflow will be triggered by every PR synchronization. 

Therefore, it would be nice to add the 'synchronize' as needed. 

If anyone has a case that `synchronize` is necessary, please feel free to share.

seokho-son: I removed "synchronize" for this time. Let's see it is necessary after some period. :)

yunkon-kim: @seokho-son 

`main` was not added to the `branches` section. Is it ok?

.github/workflows/labeler.yml
```diff
@@ -1,16 +1,19 @@
...
-  pull_request:
-    types: [opened, synchronize]
+  pull_request_target:
+    types: [opened]
     paths:
-      - 'content/**'
+      - 'content/**.md'
     branches:
-      - '**'
+      - 'dev-**'
```
seokho-son, line 8:
```suggestion
      - 'dev-**'
      - 'main'
```

.github/workflows/labeler.yml
seokho-son: @yunkon-kim Added :)

jihoon-seo (approved):
LGTM!

seokho-son:
@iamNoah1 I think this PR is ready :)

iamNoah1 (approved):
lgtm

yunkon-kim:
@seokho-son 
I think we should enable the "pull Request Labeler" workflow =)

image

PR #1317, opened 2022-09-15
add check if the term has already been declined in the past

iamNoah1:
### Describe your changes

Small but meaningful improvement. Before adding a new issue it should be also checked if the term was not already proposed but declined.


seokho-son (approved):
LGTM! Thanks @iamNoah1

jihoon-seo (approved):
LGTM!

PR #1326, opened 2022-09-20
DBaaS beyond Glossary scope?

CathPag:
I thought we decided that this term was beyond our scope. We should add a new category so we can remove terms while not totally not losing them. 

Signed-off-by: Catherine Paganini <74001907+CathPag@users.noreply.github.com>

  - [signing off](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) (ex: `git commit -s`) is to affirm that commits comply [DCO](https://wiki.linuxfoundation.org/dco).

### DIFF
--- a/content/en/database-as-a-service.md
+++ b/content/en/database-as-a-service.md
@@ -1,6 +1,6 @@
 ---
 title: Database as a Service (DBaaS)
-status: Completed
+status: Beyond Glossary scope
 category: Technology
 tags: ["", "", ""]
 ---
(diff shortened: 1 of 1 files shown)


content/en/database-as-a-service.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: Database as a Service (DBaaS)
-status: Completed
+status: Beyond Glossary scope
```
seokho-son, line 3:
```suggestion
status: Deprecated
```
How about using a simple term for this status.

jihoon-seo (approved):

seokho-son (approved):
lgtm!

PR #1355, opened 2022-10-08
correct wording for status as only english wording gets evaluated

iamNoah1:
Signed-off-by: Noah Ispas (iamNoah1) <noahispas@gmail.com>

### Describe your changes
correct wording for status as only english wording gets evaluated


seokho-son (approved):
lgtm :0

PR #1494, opened 2022-11-24
[ur] Localize 'api-gateway' in Urdu

waleed318:
Signed-off-by: Waleed Ahmed <waleedahmad318@gmail.com>

### Describe your changes
Added api-gateway.md file


Saim-Safdar (commented):
Waleed some sentences lack context when we translated them into Urdu. Can you please make sure the context is in sync plus I would love to see easy grammatic words chosen from Urdu so everyone can read and understand the concept that is written here, for me looks like we are targetting the engineering people more but I'm hopeful this translation should be the focus around non-tech people, to make it easier we need to choose an easier context and easier grammar? Let me know what do you think about it?

waleed318:
Thank you Saim, for pointing this out. Yes, you are right the main purpose of glossary is to teach non tech people difficult concepts in cloud native. 
 I will make sure context is right and will ensure the explanation of deep concept in details.

seokho-son:
Hi @waleed318 I'd like to check status of this PR. 
Actually, I hope to delete your feature branch (waleed318/glossary) which lives in the Upstream repo. 
(please avoid creating a feature branch in upstream. Every folks will download that feature branches when they fork the upstream repo.)

PR #1503, opened 2022-11-29
Add content/en/role-based access control(RBAC)

### LINKED ISSUE #1465: [New term] `Role-based access control (RBAC)`
#### What this term is about? (briefly)

Role-based access control (RBAC) is a security approach that authorizes and restricts system access to users based on their role(s) within an organization.

#### Why this term needs to be added? (briefly) 

Enterprise environments often involve multiple teams working on cloud native projects and applications on the same cloud infrastructure. When it comes to managing security for such environments, this complex weave of permissions and roles can get ugly quickly as more teams and stakeholders get involved. RBCA ensures least privilege access is maintained without making its management unwieldy.

### 2. Checklist

- [X] The suggested term fits the [cloud native definition](https://github.com/cncf/toc/blob/main/DEFINITION.md) or is essential to understanding cloud native terms. (We hope to avoid adding general or broad terminologies)
- [X] There is no existing issue yet. [(shift+click to check existing issues)](https://github.com/cncf/glossary/issues?q=label%3A%22new+term%22)
- [X] The term has not already been declined in the past. [(shift+click to check)](https://github.com/cncf/glossary/issues?q=is%3Aissue+is%3Aclosed+label%3A%22triage%2Fnot+accepted%22)
- [X] The term does not already exist in the [glossary](https://glossary.cncf.io).
- [X] I understand that maintainers will assess if the term should be part of the Glossary. (Please be patient)
- [X] I want to work on this term. (If not checked, maintainers will assign another volunteer)

### (Note)

Maintainers will assign a label after assessment procedure. 
(from `triage/awaiting` to `triage/accepted` or `triage/not accepted`)

A contributor can be assigned to this issue after the issue get the `triage/accepted` label.
The assignee needs to check [Style Guide](https://glossary.cncf.io/style-guide/).

juliafmorgado:
Signed-off-by: Julia Furst Morgado <52685951+juliafmorgado@users.noreply.github.com>

### Describe your changes
Added new term: Role-based access control (RBAC)

`resolves #1465`

### DIFF
--- a/content/en/role-based-access-control.md
+++ b/content/en/role-based-access-control.md
@@ -1,0 +1,18 @@
+---
+title: Role-Based Access Control (RBAC)
+status: Feedback Appreciated
+category: concept
+---
+
+## What it is
+
+The method of regulating access to computer or network resources based on individual users' roles within an organization.
+
+## Problem it addresses
+
+To maintain the security of an organization, it becomes increasingly important to limit unnecessary permissions based on employee roles.
+
+## How it helps
+
+It simplifies access management and enables more granular control of user actions in software applications.
+The allocation of features to the right people prevents overprivileged accounts and unauthorized access.
--- a/content/en/role-based-access-control.md
+++ b/content/en/role-based-access-control.md
@@ -18,5 +18,6 @@ Without RBAC, it is easy to lose track of who has access to what. Malicious acto
 
 ## How it helps
 
-It simplifies access management and enables more granular control of user actions in software applications.
+RBAC enables granular control over user permissions within software systems. 
+Depending on their roles, team members might be able only to view a section, change it, or have no access at all. 
 The allocation of features to the right people prevents overprivileged accounts and unauthorized access.
--- a/content/en/role-based-access-control.md
+++ b/content/en/role-based-access-control.md
@@ -7,6 +7,9 @@ category: concept
 ## What it is
 
 Role-Based Access Control (RBAC) is a method to regulate system and resource access based on individual users' roles within an organization.
+Each job function within an organization has a different role with specific permissions. 
+For instance, a marketing team member may have permission to view the deals sales has in the pipeline (*view* permission) but won't be allowed to edit them (*write* permission). 
+Sales and marketing have different permissions based on their roles, hence role-based access control. 
 
 ## Problem it addresses
 
--- a/content/en/role-based-access-control.md
+++ b/content/en/role-based-access-control.md
@@ -20,4 +20,4 @@ Without RBAC, it is easy to lose track of who has access to what. Malicious acto
 
 RBAC enables granular control over user permissions within software systems. 
 Depending on their roles, team members might be able only to view a section, change it, or have no access at all. 
-The allocation of features to the right people prevents overprivileged accounts and unauthorized access.
+This granular permission allocation allows various team members with different roles to work on the same system without risking.
--- a/content/en/role-based-access-control.md
+++ b/content/en/role-based-access-control.md
@@ -13,7 +13,8 @@ Sales and marketing have different permissions based on their roles, hence role-
 
 ## Problem it addresses
 
-To maintain the security of an organization, it becomes increasingly important to limit unnecessary permissions based on employee roles.
+Without RBAC, it is easy to lose track of who has access to what. Malicious actors are always in search of vulnerabilities to exploit and a mismanaged account with too many privileges can put an entire system at risk. 
+
 
 ## How it helps
 
--- a/content/en/role-based-access-control.md
+++ b/content/en/role-based-access-control.md
@@ -8,7 +8,7 @@ category: concept
 
 Role-Based Access Control (RBAC) is a method to regulate system and resource access based on individual users' roles within an organization.
 Each job function within an organization has a different role with specific permissions. 
-For instance, a marketing team member may have permission to view the deals sales has in the pipeline (*view* permission) but won't be allowed to edit them (*write* permission). 
+For instance, a marketing team member may have permission to view the deals sales has in the pipeline (*read* permission) but won't be allowed to edit them (*write* permission). 
 Sales and marketing have different permissions based on their roles, hence role-based access control. 
 
 ## Problem it addresses
--- a/content/en/role-based-access-control.md
+++ b/content/en/role-based-access-control.md
@@ -6,7 +6,7 @@ category: concept
 
 ## What it is
 
-The method of regulating systems and resources access based on individual users' roles within an organization.
+Role-Based Access Control (RBAC) is a method to regulate system and resource access based on individual users' roles within an organization.
 
 ## Problem it addresses
 
--- a/content/en/role-based-access-control.md
+++ b/content/en/role-based-access-control.md
@@ -6,7 +6,7 @@ category: concept
 
 ## What it is
 
-The method of regulating access to computer or network resources based on individual users' roles within an organization.
+The method of regulating systems and resources access based on individual users' roles within an organization.
 
 ## Problem it addresses
 
--- a/content/en/role-based-access-control.md
+++ b/content/en/role-based-access-control.md
@@ -8,7 +8,7 @@ category: concept
 
 Role-based access control (RBAC) is a method to regulate system and resource access based on individual user roles within an organization.
 Each job function has a specific role and permissions. 
-For instance, a marketing team member may have permission to view the deals sales has in the pipeline (*read* permission) but won't be allowed to edit them (*write* permission). 
+For instance, a marketing team member may have permission to view sales deals in the pipeline (*read* permission) but won't be allowed to edit them (*write* permission). 
 Sales and marketing have different permissions based on their roles, hence role-based access control. 
 
 ## Problem it addresses
--- a/content/en/role-based-access-control.md
+++ b/content/en/role-based-access-control.md
@@ -21,6 +21,6 @@ RBAC simplifies the process by providing predefined access control based on role
 
 ## How it helps
 
-RBAC enables granular control over user permissions within software systems. 
+RBAC provides granular control over user permissions within software systems. 
 Depending on their roles, team members might be able only to view a section, change it, or have no access at all. 
 This granular permission allocation allows various team members with different roles to work on the same system without risking.
--- a/content/en/role-based-access-control.md
+++ b/content/en/role-based-access-control.md
@@ -23,4 +23,4 @@ RBAC simplifies the process by providing predefined access control based on role
 
 RBAC provides granular control over user permissions within software systems. 
 Depending on their roles, team members can modify a section, only view it, or have no access at all. 
-This granular permission allocation allows various team members with different roles to work on the same system without risking.
+This granular permission allocation allows team members with different roles to work on the same system while minimizing risk.
--- a/content/en/role-based-access-control.md
+++ b/content/en/role-based-access-control.md
@@ -13,7 +13,8 @@ Sales and marketing have different permissions based on their roles, hence role-
 
 ## Problem it addresses
 
-Without RBAC, it is easy to lose track of who has access to what. Malicious actors are always in search of vulnerabilities to exploit and a mismanaged account with too many privileges can put an entire system at risk. 
+Individually managing access permissions to many users in an organization of various system resources and data is very complex considering that each user will have different set of resources the user can access. For instance, if there are 100,000 users and 1,000 resources in an organization, an administrator needs to check whether each user has authority and responsibility for each resource 100,000 times.
+Therefore, it is necessary to simplify the access control according to role groups based on authority and responsibility.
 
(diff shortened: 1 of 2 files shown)


iamNoah1 (approved):
@juliafmorgado I like it. Short and on point :) 

lgtm

iamNoah1:
> @juliafmorgado I like it. Short and on point :)
> 
> lgtm

ok that too fast. @juliafmorgado spellcheck fails, can you please add the words to the list :)

content/en/role-based-access-control.md
```diff
@@ -0,0 +1,18 @@
+---
+title: Role-Based Access Control (RBAC)
+status: Feedback Appreciated
+category: concept
+---
+
+## What it is
+
+The method of regulating access to computer or network resources based on individual users' roles within an organization.
```
seokho-son, line 9: Hi @juliafmorgado 

I am not sure the current resource targets (`computer or network resources`) of access control fully cover targets that can be handled by RBAC. 

For instance, RBAC can be applied to various resources (objects) in Kubernetes such as Pods or Services.

juliafmorgado: What about "regulating systems and resources access"?

seokho-son: @juliafmorgado 
Yes, it looks fine for me.

content/en/role-based-access-control.md
```diff
@@ -0,0 +1,18 @@
...
+category: concept
+---
+
+## What it is
+
+The method of regulating systems and resources access based on individual users' roles within an organization.
+
+## Problem it addresses
+
+To maintain the security of an organization, it becomes increasingly important to limit unnecessary permissions based on employee roles.
```
CathPag, line 13: This is not describing the problem. It's a statement of why RBAC is necessary. This section should focus on what would a world without RBAC look like? Ideally, we shouldn't even mention the solution, just focus on the problem.

content/en/role-based-access-control.md
```diff
@@ -0,0 +1,18 @@
+---
+title: Role-Based Access Control (RBAC)
+status: Feedback Appreciated
+category: concept
+---
+
+## What it is
+
+The method of regulating systems and resources access based on individual users' roles within an organization.
+
```
CathPag, line 10:
```suggestion
Each job function within an organization has a different role with specific permissions. 
For instance, a marketing team member may have permission to view the deals sales has in the pipeline (*view* permission) but won't be allowed to edit them (*write* permission). 
Sales and marketing have different permissions based on their roles, hence role-based access control. 

```

CathPag: I think an example would help here.
View/write permission might to be the right term. Can you help me out @iamNoah1?

content/en/role-based-access-control.md
```diff
@@ -0,0 +1,18 @@
+---
+title: Role-Based Access Control (RBAC)
+status: Feedback Appreciated
+category: concept
+---
+
+## What it is
+
+The method of regulating systems and resources access based on individual users' roles within an organization.
```
CathPag, line 9:
```suggestion
Role-Based Access Control (RBAC) is a method to regulate system and resource access based on individual users' roles within an organization.
```

CathPag (commented):
Thanks for your contribution, @juliafmorgado! 

This is a great start!  I added a few suggestions for the first section. The second needs rework as it's not focusing on the problem. I'll look at the 3rd later this week.

content/en/role-based-access-control.md
iamNoah1: @CathPag I like your edit and I think it's more read/write than view/write.

content/en/role-based-access-control.md
juliafmorgado: Thanks for the feedback Cath, I'll update it now.

content/en/role-based-access-control.md
```diff
@@ -0,0 +1,22 @@
...
+Sales and marketing have different permissions based on their roles, hence role-based access control. 
+
+## Problem it addresses
+
+Without RBAC, it is easy to lose track of who has access to what. Malicious actors are always in search of vulnerabilities to exploit and a mismanaged account with too many privileges can put an entire system at risk. 
+
+
+## How it helps
+
+It simplifies access management and enables more granular control of user actions in software applications.
```
CathPag, line 21:
```suggestion
RBAC enables granular control over user permissions within software systems. 
Depending on their roles, team members might be able only to view a section, change it, or have no access at all. 
```

content/en/role-based-access-control.md
```diff
@@ -0,0 +1,22 @@
...
+The allocation of features to the right people prevents overprivileged accounts and unauthorized access.
```
CathPag, line 22:
```suggestion
This granular permission allocation allows various team members with different roles to work on the same system without risking...
```
What does "overprivileged accounts and unauthorized access" mean concretely?

CathPag (commented):
Don't think "the problem it addresses" is correct. Will need @iamNoah1 and/or @IdealUsrname to chime in.

mfosterrox (commented):
My general concern is that the definition does not reference application control since RBAC refers to an application's access to resources within the system.

This definition should include RBAC enabling the creation of permissions for users AND workloads to act within an environment. By having hardened default RBAC controls, you can limit the ability of malicious actors or misguided operations teams from "breaking out." 

While limiting its attackers, it also lets organizations scale efficiently through grouping permissions and hardened defaults.

content/en/role-based-access-control.md
```diff
@@ -0,0 +1,22 @@
...
+## What it is
+
+Role-Based Access Control (RBAC) is a method to regulate system and resource access based on individual users' roles within an organization.
+Each job function within an organization has a different role with specific permissions. 
+For instance, a marketing team member may have permission to view the deals sales has in the pipeline (*read* permission) but won't be allowed to edit them (*write* permission). 
+Sales and marketing have different permissions based on their roles, hence role-based access control. 
+
+## Problem it addresses
+
+Without RBAC, it is easy to lose track of who has access to what. Malicious actors are always in search of vulnerabilities to exploit and a mismanaged account with too many privileges can put an entire system at risk.
```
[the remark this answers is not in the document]
iamNoah1: true, preventing malicious actors is not the main focus of RBAC. I cannot exactly say if before RBAC the solution was to have separate systems. I would more likely assume that the access levels have been defined on a User level rather than using roles.

content/en/role-based-access-control.md
iamNoah1: I would say that both statements are somehow correct. Overpriviliged would mean that users or groups have more permissions than they need, while unauthorized ones would have fewer permissions than they need?!

iamNoah1:
> My general concern is that the definition does not reference application control since RBAC refers to an application's access to resources within the system.
> 
> This definition should include RBAC enabling the creation of permissions for users AND workloads to act within an environment. By having hardened default RBAC controls, you can limit the ability of malicious actors or misguided operations teams from "breaking out."
> 
> While limiting its attackers, it also lets organizations scale efficiently through grouping permissions and hardened defaults.

+1 for adding workloads/applications and not only restricting RBAC to users/groups.

The aspect of scaling efficiently is something I never thought of before. This could be something that we should incorporate. 
We could say that before RBAC, permissions would have to be given to each user separately, which is error-prone and does not scale. Using RBAC kind of standardizes that, which results in better scaling and hardening.

CathPag:
Great! Can you add that as "committable" suggestions? It did feel a little simplified overall.

content/en/role-based-access-control.md
juliafmorgado: As Noah said, overprivileged means that users or groups have more permissions than they need. The same applies to unauthorized access; users or groups have access to systems they shouldn't have.

iamNoah1 (approved):
lgtm

content/en/role-based-access-control.md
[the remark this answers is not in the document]
iamNoah1: I guess we need some third opinion here. @seokho-son @jihoon-seo @mfosterrox what do you think?

content/en/role-based-access-control.md
```diff
@@ -0,0 +1,23 @@
...
```
seokho-son, line 16: I agree with @CathPag and @iamNoah1 
I think we'd better focus on `role` and not `access control` in the problem it address.

following is my suggestion. (please improve it if necessary)

```suggestion
Individually managing access permissions to many users in an organization of various system resources and data is very complex considering that each user will have different set of resources the user can access. For instance, if there are 100,000 users and 1,000 resources in an organization, an administrator needs to check whether each user has authority and responsibility for each resource 100,000 times.
Therefore, it is necessary to simplify the access control according to role groups based on authority and responsibility.
```

seokho-son (commented):
Added an opinion regarding the problem it address.

iamNoah1:
Thanks @juliafmorgado and sorry that it is so much back and forth going on. This is a perfect example of how we need to pay attention to all small details because we aim to be a go-to point for cloud native terms. This is a huge responsibility. However, I feel that we could now merge it. What do you think @CathPag? Some editing needed?

content/en/role-based-access-control.md
```diff
@@ -0,0 +1,24 @@
...
+## What it is
+
+Role-Based Access Control (RBAC) is a method to regulate system and resource access based on individual users' roles within an organization.
+Each job function within an organization has a different role with specific permissions. 
+For instance, a marketing team member may have permission to view the deals sales has in the pipeline (*read* permission) but won't be allowed to edit them (*write* permission). 
+Sales and marketing have different permissions based on their roles, hence role-based access control. 
+
+## Problem it addresses
+
+Individually managing access permissions to many users in an organization of various system resources and data is very complex considering that each user will have different set of resources the user can access. For instance, if there are 100,000 users and 1,000 resources in an organization, an administrator needs to check whether each user has authority and responsibility for each resource 100,000 times.
```
CathPag, line 16:
```suggestion
Individually managing access permissions of multiple users across various system resources and data can be complex.
After all, each user likely needs access to a different set of resources. 
Let's say an organization has 500 developers (users) and 300 microservices with varying permission levels (read, write, and no access). The administrator must ensure each user has the correct permissions for all 500 resources.  
```
I changed the example to a smaller group. 100,000 is a huge organization. Most people work at much smaller companies. 500 developers is still big but might feel more real. We can also decrease that number to 50 (?)

content/en/role-based-access-control.md
```diff
@@ -0,0 +1,24 @@
...
+Therefore, it is necessary to simplify the access control according to role groups based on authority and responsibility.
```
CathPag, line 17:
```suggestion
RBAC simplifies the process by providing predefined access control based on role groups.
```

content/en/role-based-access-control.md
```diff
@@ -0,0 +1,24 @@
...
+
+
+## How it helps
+
+RBAC enables granular control over user permissions within software systems.
```
CathPag, line 22:
```suggestion
RBAC provides granular control over user permissions within software systems. 
```

content/en/role-based-access-control.md
```diff
@@ -0,0 +1,24 @@
...
+## Problem it addresses
+
+Individually managing access permissions to many users in an organization of various system resources and data is very complex considering that each user will have different set of resources the user can access. For instance, if there are 100,000 users and 1,000 resources in an organization, an administrator needs to check whether each user has authority and responsibility for each resource 100,000 times.
+Therefore, it is necessary to simplify the access control according to role groups based on authority and responsibility.
+
+
+## How it helps
+
+RBAC enables granular control over user permissions within software systems. 
+Depending on their roles, team members might be able only to view a section, change it, or have no access at all.
```
CathPag, line 23:
```suggestion
Depending on their roles, team members can modify a section, only view it, or have no access at all. 
```

content/en/role-based-access-control.md
```diff
@@ -0,0 +1,24 @@
...
+
+Individually managing access permissions to many users in an organization of various system resources and data is very complex considering that each user will have different set of resources the user can access. For instance, if there are 100,000 users and 1,000 resources in an organization, an administrator needs to check whether each user has authority and responsibility for each resource 100,000 times.
+Therefore, it is necessary to simplify the access control according to role groups based on authority and responsibility.
+
+
+## How it helps
+
+RBAC enables granular control over user permissions within software systems. 
+Depending on their roles, team members might be able only to view a section, change it, or have no access at all. 
+This granular permission allocation allows various team members with different roles to work on the same system without risking.
```
CathPag, line 24:
```suggestion
This granular permission allocation allows team members with different roles to work on the same system while minimizing risk.
```

CathPag (commented):
A few small edits. What do you think, @iamNoah1?

seokho-son (approved):
LGTM. Thanks! :)

PR #1550, opened 2022-12-29
update service-mesh.md

withgaurav:
Signed-off-by: Gaurav Shukla <gaurav.shukla.gs48@gmail.com>

### Describe your changes
I have seen that there were two empty cells in the table. So I tried to fix that 


jihoon-seo:
@withgaurav Thank you for your interest and contribution, but
- please restrict your each PR to contain changes for one language only. Then the reviewers of each language will be able to review your suggestions.
- please restrict your each PR to contain changes that are relevant to each other. Please avoid concatenating irrelevant changes into one single PR.
- yes there exist some empty entities in some files' `tags` front-matter, but we don't have to remove those necessarily. We can just leave those as-is.
- if the value of `status` front-matter is something other than exact `Completed` or `completed`, then those terms will not be exposed on the CNCF Glossary website. So please don't alter the value to other than exact `Completed` or `completed`.

withgaurav:
Okay sir thank you so much but please can you guide me how can I contribute

PR #1570, opened 2023-01-01
Fix taxonomy issues

cjyabraham:
Fixes a few small issues with the current taxonomies:
- `draft: true` needs to be added to terms that are no longer "Completed" such as those that are now "Deprecated"; this removes them from the tag count on [this page](https://glossary.cncf.io/tags/) and will remove them from the google search index
- changed 2 instances of the "fundamentals" tag to be "fundamental" which is more common
- removed a "Deprecated" tag that was only used in one instance
- made terms title-case on [this page](https://glossary.cncf.io/tags/) to agree with all other instances
- reference to google_news internal template is removed since it is deprecated
- new way of getting .Path is implemented to prepare for upcoming change in Hugo

### DIFF
--- a/content/en/containers-as-a-service.md
+++ b/content/en/containers-as-a-service.md
@@ -2,6 +2,7 @@
 title: Containers as a Service
 status: Deprecated
 category: Technology
+draft: true
 tags: ["platform", "", ""]
 ---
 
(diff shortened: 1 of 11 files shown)


jihoon-seo (approved):
Preview: https://deploy-preview-1570--cncfglossary.netlify.app/tags/

LGTM!

One more thing is that it would be great to document some details about the new front-matter (`draft: true`).
(CC @CathPag @iamNoah1 @seokho-son @IdealUsrname)

seokho-son (approved):
Thanks! LGTM

PR #1612, opened 2023-01-10
feat: stale issues and PR with not activity

krol3:
### Describe your changes
Add an action to close PR without activity in the `lang/es`

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,23 @@
+name: "Stale label"
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  branches:
+    - dev-es    
+jobs:
+  stale:
+    timeout-minutes: 1
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v7
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This issue is stale because it has been labeled with inactivity.'
+        stale-pr-message: 'This PR is stale because it has been labeled with inactivity.'
+        exempt-issue-labels: 'maintainers, lang/es'
+        exempt-pr-labels: 'maintainers'
+        stale-pr-label: 'stale'
+        stale-issue-label: 'stale'
+        days-before-stale: 60
+        days-before-close: 20
+        only-labels: 'lang/es'
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,19 +1,22 @@
 name: "Stale label"
 on:
   schedule:
-    - cron: '0 0 * * *'
+    - cron: '0 0 * * *' # Runs daily at 12am UTC
   branches:
     - dev-es    
+permissions:
+  issues: write
+  pull-requests: write
 jobs:
   stale:
-    timeout-minutes: 1
+    timeout-minutes: 1 # Default: 360 
     runs-on: ubuntu-latest
     steps:
     - uses: actions/stale@v7
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-        stale-issue-message: 'This issue is stale because it has been labeled with inactivity.'
-        stale-pr-message: 'This PR is stale because it has been labeled with inactivity.'
+        stale-issue-message: 'This issue/PR is stale because it has been inactive for 60 days. This issue/PR will be closed automatically with additional 20 days of inactivity. If an update/comment occur on stale issue/PR, the stale label will be removed and the timer will restart. Closed issue/PR can be reopened if necessary.'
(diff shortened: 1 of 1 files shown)


iamNoah1:
cc @CathPag @seokho-son @jihoon-seo @IdealUsrname how do you feel about this? Im mean the localization teams could handle their contribution workflow like they want to, but I feel like closing stale PRs is not the right way to go.

CathPag:
So this is something that automates that process, right? It would be a bummer to close a stale issue that has good content but am not sure what best practices generally are to handle these things.

krol3:
> cc @CathPag @seokho-son @jihoon-seo @IdealUsrname how do you feel about this? Im mean the localization teams could handle their contribution workflow like they want to, but I feel like closing stale PRs is not the right way to go.

This is the situation when the PR is stale, because the contributor does not answer the messages or apply the suggestions. We can discuss this in the glossary meeting, what will be the best approach in this specific situation? 
cc: @electrocucaracha

electrocucaracha:
Keeping PRs open for a while could reflect a low level of participation in the project and increase the technical debt on it. Having a third party actor who does that can alleviate frictions among the community members. Besides, any PR can be re-opened any time.

CathPag:
I'm ok with what the projects leads think is best for their process :)

.github/workflows/stale.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  stale:
+    timeout-minutes: 1
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v7
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This issue is stale because it has been labeled with inactivity.'
+        stale-pr-message: 'This PR is stale because it has been labeled with inactivity.'
```
seokho-son, line 16:
```suggestion
        stale-issue-message: 'This issue/PR is stale because it has been inactive for 60 days. This issue/PR will be closed automatically with additional 20 days of inactivity. If an update/comment occur on stale issue/PR, the stale label will be removed and the timer will restart. Closed issue/PR can be reopened if necessary.'
        stale-pr-message: 'This issue/PR is stale because it has been inactive for 60 days. This issue/PR will be closed automatically with additional 20 days of inactivity. If an update/comment occur on stale issue/PR, the stale label will be removed and the timer will restart. Closed issue/PR can be reopened if necessary.'
```

I think we'd better to provide full context on the stale and automatic close policy to contributors.

seokho-son, line 16: https://github.com/actions/stale#stale-issue-message

stale-issue-message
- Required Permission: issues: write

stale-pr-message
- Required Permission: pull-requests: write

It seems these options of this workflow requires write permissions.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,23 @@
+name: "Stale label"
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  branches:
+    - dev-es    
+jobs:
+  stale:
+    timeout-minutes: 1
```
seokho-son, line 9: do we need to set this parameter ?

seokho-son (commented):
Hi @krol3 @electrocucaracha 
Thank you for this suggestion!
I think this workflow is beneficial. I also agree with testing the workflow in dev-en first. (please share the experience to the glossary community later)
To reduce the concerning point from @iamNoah1 , I think we can provide additional information regarding closed PRs so that contributors understand how it works. (commented as below)

I have some comments regarding the workflow configuration itself.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v7
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This issue is stale because it has been labeled with inactivity.'
+        stale-pr-message: 'This PR is stale because it has been labeled with inactivity.'
+        exempt-issue-labels: 'maintainers, lang/es'
+        exempt-pr-labels: 'maintainers'
+        stale-pr-label: 'stale'
```
seokho-son, line 19: - https://github.com/actions/stale#close-issue-label
- https://github.com/actions/stale#close-pr-label

looks useful for maintaining as well.

.github/workflows/stale.yml
krol3: I think, this parameter reduce the default time, and the waiting time in the job.
Check " The maximum number of minutes to let a job run before GitHub automatically cancels it. Default: 360" in https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

.github/workflows/stale.yml
krol3: Thank you @seokho-son  for your comments! I considered it, please review it

seokho-son (commented):
LGTM! looking forward a nice experience for dev-es !

@jihoon-seo could you final check this PR and merge?

ramrodo (approved):
Thanks, Carol!
/lgtm

PR #1804, opened 2023-03-10
Cloud computing update

CathPag:
### Describe your changes
Removed proprietary products and added something about public and private cloud as discussed during the maintainer meeting.

- [ x] This PR does not contain plagiarism
- [ x] I have signed off on all commits 

### DIFF
--- a/content/en/cloud-computing.md
+++ b/content/en/cloud-computing.md
@@ -7,18 +7,17 @@ tags: ["infrastructure", "fundamental", ""]
 
 ## What it is
 
-Cloud computing is a model that offers compute resources like CPU, network, and disk capabilities on-demand over the internet, allowing users to access and use computing power in a remote physical location. 
-Cloud computing can be classified into private and public cloud depending on whether the cloud infrastructure is dedicated to an organization for exclusive use or shared for open public services.
+Cloud computing offers compute resources like CPU, network, and disk capabilities on-demand over the internet, allowing users to access and use computing power in a remote physical location. 
+We generally differentiate between private and public cloud, depending on whether the cloud infrastructure is exclusively dedicated to an organization or shared for open public services.
 
 ## Problem it addresses
 
-Organizations traditionally faced two main problems when attempting to expand their use of computing power. 
-They either acquire, support, design, and pay for facilities 
-to host their physical servers and network or expand and maintain those facilities. 
-Cloud computing allows organizations to outsource some of their computing needs to another organization.
+Organizations traditionally faced two main challenges when attempting to expand computing power. 
+They could either acquire, support, and design (new) facilities to host their physical servers and network or expand and maintain existing ones. 
+Cloud computing solves that challenge by allowing organizations to outsource some of their computing needs. 
 
 ## How it helps
 
 Cloud providers allow organizations to rent compute resources on-demand and pay for usage, delivering two key benefits.
-Organizations can focus on their product or service without waiting, planning, and spending resources on new physical infrastructure. They can simply [scale](/scalability/) as needed and on-demand.
+First, organizations can focus on their product or service without waiting, planning, and spending resources on new physical infrastructure. And second, they can simply [scale](/scalability/) on-demand as needed.
 Cloud computing allows organizations to adopt as much or as little infrastructure as they need.


content/en/cloud-computing.md
```diff
@@ -7,21 +7,20 @@ tags: ["infrastructure", "fundamental", ""]
...
 ## What it is
 
-Cloud computing is a model that offers compute resources like CPU, network, and disk capabilities on-demand over the internet. 
-Cloud computing gives users the ability to access and use computing power in a remote physical location. 
-Cloud providers like AWS, GCP, Azure, DigitalOcean, and others all offer third parties 
-the ability to rent access to compute resources in multiple geographic locations.
+Cloud computing is a model that offers compute resources like CPU, network, and disk capabilities on-demand over the internet, allowing users to access and use computing power in a remote physical location. 
+Cloud providers offer third parties the ability to rent access to compute resources in multiple geographic locations.
+Users rent these compute resources in a public or private virtual cloud (PVC).
+In a public cloud, users share physical hardware with other users, whereas in a PVC, the user rents an entire server.
```
seokho-son, line 13: I think `Cloud providers offer third parties the ability to rent access to compute resources in multiple geographic locations.` is a duplicated sentence with the sentence above. (`, allowing users to access and use computing power in a remote physical location.)

Also, I think we don't have to introduce `private virtual cloud (PVC)` concept here. Looks like too details. (By the way, I guess PVC is typo of VPC. VPC does not directly mean a Private cloud. It can be served by a Public cloud service.)

What about just mentioning public cloud and private cloud simply.

```suggestion
Cloud computing can be classified into public cloud and private cloud depending on whether the cloud computing-based services are open to the outside.
```

seokho-son (commented):
Hi @CathPag 
Thank you for this update. I have a comment.

content/en/cloud-computing.md
iamNoah1: Thanks, @seokho-son I was also not happy with the introduction of a private virtual cloud but had no good suggestions. So I like the suggestion :)

iamNoah1 (changes requested):
Just that suggestion from @seokho-son. Looks good otherwise :)

content/en/cloud-computing.md
```diff
@@ -7,21 +7,18 @@ tags: ["infrastructure", "fundamental", ""]
 
 ## What it is
 
-Cloud computing is a model that offers compute resources like CPU, network, and disk capabilities on-demand over the internet. 
-Cloud computing gives users the ability to access and use computing power in a remote physical location. 
-Cloud providers like AWS, GCP, Azure, DigitalOcean, and others all offer third parties 
-the ability to rent access to compute resources in multiple geographic locations.
+Cloud computing is a model that offers compute resources like CPU, network, and disk capabilities on-demand over the internet, allowing users to access and use computing power in a remote physical location. 
+Cloud computing can be classified into public and private cloud depending on whether the cloud services are open to the outside.
```
CathPag, line 11: Not sure about "whether cloud services are open to the outside (world)." Isn't it about whether you share actual hardware or not? Public cloud: you have your own virtual resources but share the hardware with other cloud customers. Private cloud: your virtualized environment runs on it's own dedicated machine. (cc @seokho-son @iamNoah1)

seokho-son:
> Hmm...I thought that was the conclusion of our last meeting. Instead of adding a VPC term, we reference VPC and public cloud in this entry. I was just implementing our decision. Should we skip it here AND not have a VPC term?
> 
> Meeting note: New term suggestion Public Cloud issues/1737 (related with ongoing effort for VPC) → Agreed not to include, and mentioning it in cloud computing term.

Hi @CathPag :)

I thought the conclusion of our last meeting is.. Instead of adding `Public cloud` as a new term, just adding short description about `Public cloud` and `Private cloud` in the existing `Cloud computing` term so that the VPC term (ongoing work) can reference public cloud from `Cloud computing`.
I guess we didn't make another decision for the VPC term itself.

content/en/cloud-computing.md
seokho-son: @CathPag 
In my understanding, sharing underlying hardware of virtual resources is not the factor that differentiates a Public cloud and a Private cloud.

Dedicating or sharing hardware in a cloud computing is another decision point regardless of a Public/Private cloud. For instance, we can request a virtual machine using a dedicated hardware in AWS (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/dedicated-instance.html). 

The following is definition from [NIST](https://nvlpubs.nist.gov/nistpubs/legacy/sp/nistspecialpublication800-145.pdf)

> Private cloud. The cloud infrastructure is provisioned for exclusive use by a single organization
> comprising multiple consumers (e.g., business units). It may be owned, managed, and
> operated by the organization, a third party, or some combination of them, and it may exist
> on or off premises.

> Public cloud. The cloud infrastructure is provisioned for open use by the general public. It may be
> owned, managed, and operated by a business, academic, or government organization, or
> some combination of them. It exists on the premises of the cloud provider.

CathPag: Hmm...this sounds to me like the (real, not virtualized) hardware is dedicated: "The cloud infrastructure is provisioned for exclusive use..." Where I was wrong is assuming it was owned by a third party exclusively. According to NIST's definition, an organization can have its own private cloud (not really sure how that differs from a regular data center, though, but that's a different topic.)

The AWS example doesn't really apply because it's a VPC, so virtualized hardware, not real hardware.

seokho-son: @CathPag 
I guess my confusion is from the dedication of hardware. Both private cloud and public cloud are (generally) based on virtualization technology (making VMs from Physical hardware). There is a hardware dedication concept in virtualization technology. We can dedicate specific hardware to provision VMs or VPCs. 

I agree with "the (real, not virtualized) hardware is dedicated" for exclusive use in a private cloud. 
How about this ?
```suggestion
Cloud computing can be classified into private and public cloud depending on whether the cloud infrastructure is dedicated to an organization for exclusive use or shared for open public services.
```


Regarding
> (not really sure how that differs from a regular data center, though, but that's a different topic.)

Let me share my understanding. A data center is a computing infrastructure consists of huge computing hardware. A data center can be infrastructure of a cloud (i.e., cloud data center), but not necessarily (i.e., on-premises data center).

iamNoah1 (approved):
I committed some suggestions from @seokho-son so that we can move forward here. We can come back to the term if we think that it is not accurate enough.

seokho-son (approved):
LGTM! Thanks @CathPag

PR #1884, opened 2023-04-04
Fix dev-es stale GH workflow

electrocucaracha:
### Describe your changes
This change fixes the syntax of stale workflow, removing the double quotes from the name and upgrades the version number of the task

NA

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,43 @@
+---
+# SPDX-license-identifier: Apache-2.0
+##############################################################################
+# Copyright (c) 2023
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Apache License, Version 2.0
+# which accompanies this distribution, and is available at
+# http://www.apache.org/licenses/LICENSE-2.0
+##############################################################################
+
+name: Stale label
+# yamllint disable-line rule:truthy
+on:
+  schedule:
+    - cron: '0 0 * * *' # Runs daily at 12am UTC
+  push:
+    branches:
+      - dev-es
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale:
+    name: Run stale task to label inactive issues
+    timeout-minutes: 1 # Default: 360 
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v8.0.0
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


krol3:
@electrocucaracha @ramrodo  I was testing this action, after we merge in to the main branch, the last (before)  will work fine.  I dont think we need this pr to fix it. It was your suggestion @ramrodo

.github/workflows/stale.yml
```diff
@@ -1,28 +1,43 @@
-name: "Stale label"
+---
+# SPDX-license-identifier: Apache-2.0
```
krol3, line 2: Do we need of these generated comments? my suggestion is a clean yml file

.github/workflows/stale.yml
```diff
@@ -1,28 +1,43 @@
...
-        exempt-pr-labels: 'maintainers'
-        stale-pr-label: 'stale'
-        stale-issue-label: 'stale'
-        days-before-stale: 60
-        days-before-close: 20
-        only-labels: 'lang/es'
-        close-issue-label: 'close/stale'
-        close-pr-label: 'close/stale' # Add the label for mainteiners track the close pr labels.
\ No newline at end of file
+      - uses: actions/stale@v8.0.0
```
krol3, line 30: it was tested with this new version this action?

.github/workflows/stale.yml
electrocucaracha: IMHO, it's a good practice to have headers on files, those can eventually help on SBOM adoption. The Spanish spellchecker contains the same header

.github/workflows/stale.yml
electrocucaracha: I tried to test this locally with `act`  tool but requires some env vars. Besides using the latest version released, this semver format is validated by dependabot

ramrodo (approved):

seokho-son:
I guess this workflow needs to be applied to main (base) branch. :)
I think we can apply this PR to main branch first, and open another PR to update dev-es.

.github/workflows/stale.yml
krol3: ok!! let's merge it!

krol3 (approved):
LGTM

electrocucaracha:
> I guess this workflow needs to be applied to main (base) branch. :) I think we can apply this PR to main branch first, and open another PR to update dev-es.

This file doesn't exist in main, so I was thinking to fix it first in `dev-es` and submit other PR to sync changes with `main` later.

seokho-son (approved):
Currently, this PR is targeting to the main branch. :)
Anyway, LGTM as is.

PR #1889, opened 2023-04-04
fix: action stale

krol3:
### Describe your changes
Fix the stale action
Testing here : https://github.com/krol3/glossary/actions/runs/4610611411


electrocucaracha:
@krol3 shouldn't be better to include this change on #1884 ?

krol3:
Closing in favor of https://github.com/cncf/glossary/pull/1884

PR #1979, opened 2023-04-23
Tweaks to footer layout

cjyabraham:
Makes footer consistent with all other CNCF subsites.


jihoon-seo (approved):
[As-is]
image
image

---

[To-be] ([Preview](https://deploy-preview-1979--cncfglossary.netlify.app/))
image
image

We can see that gaps are shrinked in overall.
LGTM 😊

seokho-son (approved):
lgtm!

Issue #2143, opened 2023-06-14
[Update term] `Virtualization` for `En`


guillaumebernard84:
#### Why this term needs to be updated? (briefly)
- I think the last sentence from the "How it helps" paragraph doesn't make sense: "_VMs also enable users to speed up the time to get a new virtual computer_".  Without VMs, you cannot get a new virtual computer at all. We could fix this sentence, but then it would be redundant with the previous one in the same paragraph. I sugest to remove it. If you agree with this change, I can submit a PR.  

### 2. Checklist

- [X] I understand that maintainers will confirm if the term should be updated. (Please be patient)
- [X] I want to work on this suggestion and confirm that I meet the criteria as outlined in "[who can contribute](https://glossary.cncf.io/contribute/#who-can-contribute)".. (If not checked, maintainers will assign another volunteer)

### (Note)

Maintainers will assign a label after assessment procedure. 
(from `triage/awaiting` to `triage/accepted` or `triage/not accepted`)

A contributor can be assigned to this issue after the issue get the `triage/accepted` label.
The assignee needs to check [Style Guide](https://glossary.cncf.io/style-guide/).

CathPag:
Thanks, @guillaumebernard84. Please go ahead and submit a PR. 
For small changes like this, feel free to submit a PR directly. Even if the change is rejected, it might be a similar amount of work but a much faster process overall.

guillaumebernard84:
Thank you, I just did that. Should I close this issue right now?

PR #2189, opened 2023-06-30
[EN] update Virtualization

guillaumebernard84:
### Describe your changes
I think the last sentence from the "How it helps" paragraph doesn't make sense: "VMs also enable users to speed up the time to get a new virtual computer". Without VMs, you cannot get a new virtual computer at all. We could fix this sentence, but then it would be redundant with the previous one in the same paragraph. I suggest to remove it.

Fixes https://github.com/cncf/glossary/issues/2143


nate-double-u:
Thanks for this @guillaumebernard84, I agree this sentence is unneeded.

nate-double-u (approved):

nate-double-u:
I just updated the description to make sure this PR was linked to the issue.

CathPag (approved):
LGTM

PR #2190, opened 2023-07-02
Enable check outdated gh action in dev-es

electrocucaracha:
### Describe your changes
This change fixes some types and enables the Check outdated terms in the `dev-es` branch

NA

### DIFF
--- a/.github/workflows/check-outdated-content.yaml
+++ b/.github/workflows/check-outdated-content.yaml
@@ -39,31 +39,26 @@ jobs:
         # Set output direcory
         OUTPUT_DIR="./outdated"
         
-        # Set L10n directory and code
+        # Set L10n code
+        L10N_CODE="${L10N_BRANCH//dev-/}"
+        echo "(DEBUG) L10N Code: ${L10N_CODE}"
+
+        # Set L10n directory
         case "${L10N_BRANCH}" in
           dev-pt)
-          L10N_DIR="content/pt-br/"
-          L10N_CODE="pt-br"
+          L10N_DIR="./content/pt-br/"
           ;;
-
           dev-zh)
-          L10N_DIR="content/zh-cn/"
-          L10N_CODE="zh-cn"
+          L10N_DIR="./content/zh-cn/"
           ;;
-
           dev-tw)
-          L10N_DIR="content/zh-tw/"
-          L10N_CODE="zh-tw"
+          L10N_DIR="./content/zh-tw/"
           ;;
-
           *)
-          L10N_CODE="${L10N_BRANCH//dev-/}"
           L10N_DIR="./content/${L10N_CODE}/"
           ;;
         esac
-        
         echo "(DEBUG) L10N Directory: ${L10N_DIR}"
-        echo "(DEBUG) L10N Code: ${L10N_CODE}"
         
         # Set L10N_DIR, L10N_CODE, and OUTPUT_DIR as environment variables
         # Ref: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable
@@ -106,7 +101,6 @@ jobs:
         # The old branch can be 'upstream/dev-ko'
         OLD_BRANCH="origin/${{github.base_ref}}"
         echo "(DEBUG) OLD_BRANCH: ${OLD_BRANCH}"
-        echo "(DUBUG) OLD_BRANCH: ${OLD_BRANCH}"
 
         L10N_INFO_JSON=$(cat <<EOF
         {
--- a/.github/workflows/check-outdated-content.yaml
+++ b/.github/workflows/check-outdated-content.yaml
@@ -43,12 +43,17 @@ jobs:
         case "${L10N_BRANCH}" in
           dev-pt)
           L10N_DIR="content/pt-br/"
-          L10N_CODE="pt"
+          L10N_CODE="pt-br"
           ;;
 
           dev-zh)
           L10N_DIR="content/zh-cn/"
-          L10N_CODE="zh"
+          L10N_CODE="zh-cn"
+          ;;
+
+          dev-tw)
+          L10N_DIR="content/zh-tw/"
+          L10N_CODE="zh-tw"
           ;;
 
           *)
--- a/.github/workflows/check-outdated-content.yaml
+++ b/.github/workflows/check-outdated-content.yaml
@@ -4,7 +4,8 @@ name: Check outdated content
 on: 
   pull_request:
     branches:
-      - 'dev-ko' # add other branches or use wildcard 'dev-**'
+      - 'dev-ko'
+      - 'dev-es'
     paths:
       - 'content/en/**.md'
 
@@ -12,10 +13,10 @@ jobs:
   check-outdated-content:
     name: Check outdated content
 
-    # if: contains(fromJSON('["dev-ko", "dev-xx"]'), github.base_ref) 
+    if: contains(fromJSON('["dev-ko","dev-es"]'), github.base_ref)
     # Ref: https://docs.github.com/en/actions/learn-github-actions/expressions
 
-    if: github.base_ref == 'dev-ko'
+    #if: github.base_ref == 'dev-ko'
     
     # Condition to run this workflow on the upstream repository
     #if: github.repository == 'cncf/glossary'
@@ -40,15 +41,20 @@ jobs:
         
         # Set L10n directory and code
         case "${L10N_BRANCH}" in
-          dev-ko)          
-          L10N_DIR="content/ko/"
-          L10N_CODE="ko"
+          dev-pt)
+          L10N_DIR="content/pt-br/"
+          L10N_CODE="pt"
           ;;
-          
-          #dev-pt)
-          #L10N_DIR="content/pt-br/"
-          #L10N_CODE="pt"
-          #;;
+
+          dev-zh)
+          L10N_DIR="content/zh-cn/"
+          L10N_CODE="zh"
+          ;;
+
+          *)
+          L10N_CODE="${L10N_BRANCH//dev-/}"
+          L10N_DIR="./content/${L10N_CODE}/"
+          ;;
         esac
         
         echo "(DEBUG) L10N Directory: ${L10N_DIR}"
@@ -89,11 +95,12 @@ jobs:
         # Get the lastest branch name from 'GITHUB_REF' 
         # The latest branch can be 'upstream/main' or 'forked/dev-ko' (rebased)
         LATEST_BRANCH=${GITHUB_REF#refs/}
-        echo "(DUBUG) LATEST_BRANCH: ${LATEST_BRANCH}"
+        echo "(DEBUG) LATEST_BRANCH: ${LATEST_BRANCH}"
         
         # Get the old branch from 'github.base_ref' 
         # The old branch can be 'upstream/dev-ko'
         OLD_BRANCH="origin/${{github.base_ref}}"
+        echo "(DEBUG) OLD_BRANCH: ${OLD_BRANCH}"
         echo "(DUBUG) OLD_BRANCH: ${OLD_BRANCH}"
 
         L10N_INFO_JSON=$(cat <<EOF
@@ -144,7 +151,7 @@ jobs:
               fi
 
             else
-              echo "(DEBUG) ${FILE_PATH} dose not exist."
+              echo "(DEBUG) ${FILE_PATH} does not exist."
               # File dose not exist (e.g, changed, renamed or removed)
               echo "Could not find ${FILE_PATH} in content/en/" > ${OUTPUT_DIR}/${FILE_PATH}
               echo "Need to check if it has been changed, renamed or removed" >> ${OUTPUT_DIR}/${FILE_PATH}


.github/workflows/check-outdated-terms.yaml
```diff
@@ -61,7 +52,7 @@ jobs:
         echo "OUTPUT_DIR=${OUTPUT_DIR}" >> $GITHUB_ENV
     
     - name: Checkout 
-      uses: actions/checkout@v3
+      uses: actions/checkout@v3.5.3
```
jihoon-seo, line 55: I personally prefer not to specify minor and patch versions for GitHub Action. Specifying those will incur relatively frequent PRs by bot to bump minor and patch versions every time the newer version comes.
In contrast, not specifying minor and patch versions and specifying major version only intentionally will
- make GitHub workflow pull the latest version (e.g., 3.5.3) of specified major version, and
- cause fewer PRs by bot, only when the new major version has been released.

You can see the PRs by bot to automatically bump the version of GitHub Actions at:
https://github.com/cncf/glossary/pulls?q=is%3Apr+is%3Aclosed+bump

What's your opinion? 😊

electrocucaracha: My preference is `Version pinning`. It's a best practice [for dependency management](https://cloud.google.com/blog/topics/developers-practitioners/best-practices-dependency-management) that facilitates the troubleshooting. Several tools recommend this approach, like [Hadolint](https://github.com/hadolint/hadolint/wiki/DL3008) with Dockerfiles. I agree, that can be a like bit noisy, and sometimes makes a [bot](https://docs.github.com/en/code-security/dependabot/working-with-dependabot) the top contributor of the project.

jihoon-seo: @electrocucaracha We discussed about this in the monthly meeting yesterday, and reached a conclusion:
yes it's important to pin the specific version for dependency management and so on, but we maintainers prefer not to specify minor and patch versions in this particular case, because of some reasons, where part of them might be:
- the workflows configured in this repo are somewhat auxiliary ones, so we do not need very strict dependency management.
- the GitHub Actions that we are using seem to not be likely to cause (crucial) trouble.
- as you said it might be noisy.. 😅

And these reasons might apply to our case only, so it will be encouraged to pin the version for other general cases.

PTAL!

seokho-son:
Hi @electrocucaracha,

Please note that the script applied for Korean is experimental and not perfect! :)

Anyway, in order to apply the script to another dev branch, I believe the configuration should be in the main branch (Github workflow uses configurations in main - default branch). So, merging this PR (target: dev-ex) would not be effective.

Why don't we use the main branch for this work?

seokho-son:
Hi @yunkon-kim,
Could you review this PR as the original author of the script? :)
Are there any concerns?

.github/workflows/check-outdated-terms.yaml
```diff
@@ -39,17 +39,8 @@ jobs:
...
-          L10N_CODE="ko"
-          ;;
-          
-          #dev-pt)
-          #L10N_DIR="./content/pt-br/"
-          #L10N_CODE="pt"
-          #;;
-        esac
+        L10N_CODE="${L10N_BRANCH//dev-/}"
+        L10N_DIR="./content/${L10N_CODE}/"
```
yunkon-kim, line 43: Hi @electrocucaracha 

I'm glad you're interested in this workflow even though it's being developed and tested :-)

According to my analysis of dev branches, code, and directory, **most localization teams use double-digit characters except in Portuguese and traditional Chinese.**

Let me explain this with examples.
- In Korean localization, `ko` is used in all parts, such as **`dev-ko`, `ko`, and `./content/ko`**.
- But, in Portuguese localization, `pt` and `pt-br` are used, such as **`dev-pt`, `pt`, and `./content/pt-br`**.

Therefore, it seems necessary to distinguish between the localization directory and code like the previous script.

Could you consider this and revise it?

yunkon-kim (changes requested):

.github/workflows/check-outdated-terms.yaml
electrocucaracha: Yeah, I noticed this exception on Portuguese. The Spanish team has consolidated efforts (**`es-mx`**, **`es-ar`**, **`es-es`**, etc.) into a single language collaborative effort. We believe this approach can provide us better short and long term benefits. 

Anyway, I can modify the code to consider that exception and enable this workflow for the rest of the branches or, leave it in this way, enabling branch by branch and try to address that country exception.

yunkon-kim: @electrocucaracha 

The Spanish team's approach looks great to me. I also like the concise code you suggested.

However, I prefer the first of the two options you suggested.
This is because the deleted code was designed to reflect the current status of all other teams.

```diff
        # Set L10n directory and code
-        case "${L10N_BRANCH}" in
-          dev-ko)          
-          L10N_DIR="./content/ko/"
-          L10N_CODE="ko"
-          ;;
-          
-          #dev-pt)
-          #L10N_DIR="./content/pt-br/"
-          #L10N_CODE="pt"
-          #;;
-        esac
+        L10N_CODE="${L10N_BRANCH//dev-/}"
+        L10N_DIR="./content/${L10N_CODE}/"
```

.github/workflows/check-outdated-content.yaml
```diff
@@ -4,7 +4,7 @@ name: Check outdated content
 on: 
   pull_request:
     branches:
-      - 'dev-ko' # add other branches or use wildcard 'dev-**'
+      - 'dev-*'
```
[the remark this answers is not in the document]
yunkon-kim: Another is related to lines 15 to 18.

Currently, this workflow runs only for the 'dev-ko'. So I would need you to change lines 15 to 18 like I have included below.

```diff
-    if: contains(fromJSON('["dev-ko", "dev-xx"]'), github.base_ref) 
+    if: contains(fromJSON('["dev-ko", "dev-es"]'), github.base_ref) 
    # Ref: https://docs.github.com/en/actions/learn-github-actions/expressions
-     if: github.base_ref == 'dev-ko'
+    # if: github.base_ref == 'dev-ko'
```

yunkon-kim (changes requested):

yunkon-kim (approved):
LGTM!

yunkon-kim:
@jihoon-seo PTAL

.github/workflows/check-outdated-content.yaml
```diff
@@ -40,15 +41,20 @@ jobs:
...
+          L10N_DIR="content/pt-br/"
+          L10N_CODE="pt"
+          ;;
+
+          dev-zh)
+          L10N_DIR="content/zh-cn/"
+          L10N_CODE="zh"
+          ;;
+
+          *)
```
jihoon-seo, line 54: Hmm, my suggestion is as follows:

```suggestion
          dev-pt)
          L10N_DIR="content/pt-br/"
          L10N_CODE="pt-br"
          ;;

          dev-zh)
          L10N_DIR="content/zh-cn/"
          L10N_CODE="zh-cn"
          ;;

          dev-tw)
          L10N_DIR="content/zh-tw/"
          L10N_CODE="zh-tw"
          ;;

          *)
```

By this, we can set `L10N_CODE` as `zh-cn` for `dev-zh` branch, and `zh-tw` for `dev-tw` branch.
IMO this will make the [name of GitHub workflow step](https://github.com/electrocucaracha/glossary/blob/enable_gh_check_versions/.github/workflows/check-outdated-content.yaml#L167) and the [log](https://github.com/electrocucaracha/glossary/blob/enable_gh_check_versions/.github/workflows/check-outdated-content.yaml#L109) more clear and easily identifiable.. 😊

yunkon-kim:
Hi @jihoon-seo, @electrocucaracha 

Thank you for the valuable comments and quick updates.

But I think we need to fix it once more 😅 (I'm gonna make a suggestion.)

Let me explain why.

As the below table shows, `L10N_CODE` was derived from **branches** and **labels**. It is represented by two characters.
`L10N_DIR` was derived from **directories** for each language.

Some codes and directories don't match, so I had to differentiate between the two.

Please see the analysis of labels, branches, and directories in the Glossary 😃 

| Language | Label | Branch | Directory | L10N_CODE | L10N_DIR | Note |
|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
| Arabic | lang/ar | dev-ar | content/ar | ar | content/ar/ |  |
| Bengali | lang/bn | dev-bn | content/bn | bn | content/bn/ |  |
| German | lang/de | dev-de | content/de | de | content/de/ |  |
| English | lang/en | main | content/en | en | content/en/ | It's for i18n :-D |
| Spanish | lang/es | dev-es | content/es | es | content/es/ |  |
| French | lang/fr | dev-fr | content/fr | fr | content/fr/ |  |
| Hindi | lang/hi | dev-hi | content/hi | hi | content/hi/ |  |
| Italian | lang/it | dev-it | content/it | it | content/it/ |  |
| Japanese | lang/ja | dev-ja | content/ja | ja | content/ja/ |  |
| Korean | lang/ko | dev-ko | content/ko | ko | content/ko/ |  |
| Portuguese | lang/pt | dev-pt | content/pt-br | pt | content/pt-br/ |  |
| Russian | lang/ru | dev-ru | content/ru | ru | content/ru/ |  |
| Turkish | lang/tr | dev-tr | content/tr | tr | content/tr/ |  |
| Traditional Chinese | lang/tw | dev-tw | content/zh-tw | tw | content/zh-tw/ |  |
| Urdu | lang/ur | dev-ur | content/ur | ur | content/ur/ |  |
| Chinese | lang/zh | dev-zh | content/zh-cn | zh | content/zh-cn/ |  |

.github/workflows/check-outdated-content.yaml
yunkon-kim: Related to the [name of GitHub workflow step](https://github.com/electrocucaracha/glossary/blob/enable_gh_check_versions/.github/workflows/check-outdated-content.yaml#L167), this is the directory where we store diff results between files. 

This is the `${OUTPUT_DIR}` 😅 
```
OUTPUT_DIR="./outdated"
```

I think it can be classifiable by L10N_CODE in the name section below.
```
    - name: Upload output
      uses: actions/upload-artifact@v3
      with:
        name: ${{ env.L10N_CODE }}-outdated-checking-result
        path: ${{ env.OUTPUT_DIR }}/
```


The [log](https://github.com/electrocucaracha/glossary/blob/enable_gh_check_versions/.github/workflows/check-outdated-content.yaml#L109) part is redundant, so we need to remove that area.

yunkon-kim (commented):
Here's my suggestion.

.github/workflows/check-outdated-content.yaml
```diff
@@ -40,15 +41,25 @@ jobs:
...
+          ;;
+
+          dev-tw)
+          L10N_DIR="content/zh-tw/"
+          L10N_CODE="zh-tw"
+          ;;
+
+          *)
+          L10N_CODE="${L10N_BRANCH//dev-/}"
+          L10N_DIR="./content/${L10N_CODE}/"
```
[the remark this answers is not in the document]
electrocucaracha: @yunkon-kim thanks for catching that, so I think in that case `L10N_CODE` codes can be gotten directly from `L10N_BRANCH`, in other words it's possible to use `L10N_CODE="${L10N_BRANCH//dev-/}"` before the switch/case instruction as long as `main` branch is not enabled.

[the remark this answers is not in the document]
yunkon-kim: @electrocucaracha I agree with your suggestion! 

The switch/case had been programmed before I added the condition (`if: github.base_ref == 'dev-ko'`) as in [line 19](https://github.com/cncf/glossary/blob/118824c740fe715644fb950724b7b61d7f6af47a/.github/workflows/check-outdated-content.yaml#L19). Now we have improved that as in [line 16](https://github.com/cncf/glossary/blob/118824c740fe715644fb950724b7b61d7f6af47a/.github/workflows/check-outdated-content.yaml#L16) `if: contains(fromJSON('["dev-ko","dev-es"]'), github.base_ref)`

So, I think we don't need to consider `main`. 

Please improve this according to your suggestions.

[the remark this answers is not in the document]
electrocucaracha: So the condition needs to be something like `if: github.base_ref != 'main'`?

yunkon-kim:
@electrocucaracha 

Oh~ you already improved the code. Your improvement is fine for me. Nonetheless, I should share one thing. 

Related to the L10N_DIR part, I like the previous code because it has better readability.

electrocucaracha:
> @electrocucaracha
> 
> Oh~ you already improved the code. Your improvement is fine for me. Nonetheless, I should share one thing.
> 
> Related to the L10N_DIR part, I like the previous code because it has better readability.

What about using a hash table?

```bash
        declare -A codes=( ["dev-pt"]="pt-br" ["dev-zh"]="zh-cn" ["dev-tw"]="zh-tw")
        [ ! -v ${codes[$L10N_BRANCH]} ] && L10N_DIR="./content/${codes[$L10N_BRANCH]}/" || L10N_DIR="./content/${L10N_CODE}/"
        echo "(DEBUG) L10N Directory: ${L10N_DIR}"
```

yunkon-kim:
@electrocucaracha 

Your code is pretty efficient. Meanwhile, we have to consider that this workflow will gradually be used by all localization teams. In the future, other folks may need to improve this workflow.

Therefore, I would like to suggest that it would be better to prioritize code readability as in the previous code.

yunkon-kim:
Hi @jihoon-seo 

With meaningful approaches and discussions, this PR is mature enough to merge. Could you please take a look finally? :-D

jihoon-seo (approved):
LGTM from my side!
(Hash table suggested by @electrocucaracha also LGTM.)

iamNoah1:
hold for @seokho-son's review

seokho-son (approved):
LGTM! :)
Thanks All.

Issue #2242, opened 2023-07-19
Change "What it is" heading in definition template


nate-double-u:
The heading "What it is" would be more grammatically correct as: "What is it?"

This update would involve changing [the template in the Style Guide](https://glossary.cncf.io/style-guide/#definition-template) and updating all the en language terms that have been added so far.
```[tasklist]
### Tasks
```

rolfedh:
- + 1. I like the proposed changes. Generally, titles that ask a question spur readers' curiosity, engagement, and retention. 
- Although we _could_ express some of the other section titles as questions, but I prefer not to. Variation is also important for retaining interest.

iamNoah1:
after long discussion about this topic among the maintainers, the consensus is to remove the what it is header. see #2575

PR #2243, opened 2023-07-19
Updating heading in new word template "What is it?"

nate-double-u:
resolves: #2242 

- Updating new word template heading to read "What is it?" 
- Updating all existing en pages to reflect new template

### DIFF
--- a/content/en/_TEMPLATE.md
+++ b/content/en/_TEMPLATE.md
@@ -4,7 +4,7 @@ status: Feedback Appreciated
 category: concept
 ---
 
-## What it is
+## What is it?
 
 Quick summary of the concept and what it is.
 
--- a/content/en/agile-software-development.md
+++ b/content/en/agile-software-development.md
@@ -5,7 +5,7 @@ category: concept
 tags: ["methodology", "", ""]
 ---
 
-## What it is
+## What is it?
 
 A set of practices that emphasize iterative development cycles and self-organizing teams. 
 In contrast to waterfall-like projects where value is generated only at the very end of a project, 
--- a/content/en/api-gateway.md
+++ b/content/en/api-gateway.md
@@ -5,7 +5,7 @@ category: technology
 tags: ["networking", "", ""]
 ---
 
-## What it is
+## What is it?
 
 An [API](/application-programming-interface/) gateway is a tool that 
 aggregates unique application APIs, making them all available in one place. 
--- a/content/en/application-programming-interface.md
+++ b/content/en/application-programming-interface.md
@@ -5,7 +5,7 @@ category: technology
 tags: ["architecture", "fundamental", ""]
 ---
 
-## What it is
+## What is it?
 
 An API is a way for computer programs to interact with each other. 
 Just as humans interact with a website via a web page, an API allows computer programs to interact with each other. 
--- a/content/en/bare-metal-machine.md
+++ b/content/en/bare-metal-machine.md
@@ -5,7 +5,7 @@ category: technology
 tags: ["infrastructure", "", ""]
 ---
 
-## What it is
+## What is it?
 
 Bare metal refers to a physical computer, more specifically a server, that has one, and only one, operating system. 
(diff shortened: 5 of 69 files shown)


CathPag (approved):
LGTM

seokho-son:
Hi @nate-double-u :)
While I'm not very proficient in English grammar, I'd like to check if changing "What it is" to a question form like "What is it?" could introduce a bit of inconsistency with the form of the other headers ("Problem it addresses" and "How it helps"). 

What it is -> What is it?
Problem it addresses -> What problem does it address?
How it helps -> How does it help?

nate-double-u:
> Hi @nate-double-u :) While I'm not very proficient in English grammar, I'd like to check if changing "What it is" to a question form like "What is it?" could introduce a bit of inconsistency with the form of the other headers ("Problem it addresses" and "How it helps").
> 
> What it is -> What is it? Problem it addresses -> What problem does it address? How it helps -> How does it help?

I think the inconsistency is OK in this case (tho, am happy for more comments here). The way I read it, the first is a question—what is the thing we're defining—then, once explained or introduced, the other two section titles refer back to that "it," and are more declarative.

I've been looking at it more, and I think both "What it is" and "What is it?" are both correct, and I'm leaning towards this actually being more of a style choice than a grammar issue.

So the question in #2242, really is a question of preference.

chalin:
@nate-double-u ask for some feedback. We're treading in a **grey zone** here, so this might be **bike-shedding** -- but here's my take anyway.

---

(0) I agree that "What it is" feels a bit off, but in the context of other headings, I agree with @seokho-son that there is a uniformity that makes sense.

(1) I think that a fix would be to **drop the "What it is" heading altogether**: implicitly, each page has a H1 header corresponding to the term being defined, so we don't need an H2 to immediately follow it. For example, consider https://glossary.cncf.io/data-center/ without that first H2:

| Before | After |
|--------|--------|
| image | image

(2) I also agree with @seokho-son that if y'all choose to change the first H2 title, then it makes sense to make them all questions.

---

That being said, my vote would be for the following, in **order of decreasing preference**, using the numbers given above:

- (1) drop "What it is" headings
- (0) Status quo -- leave the headings as they are
- (2) Update all heading titles to be questions

nate-double-u:
Thanks for this @chalin, I think I agree, option (1), dropping the "What it is" headings, is a good solution.

iamNoah1:
I have no strong opinion, so I am fine with the suggestion

seokho-son:
Hi @nate-double-u @chalin 
Thank you for the enhancement idea!

I think we could consider removing the first header as suggested, assuming it provides enough benefits, even though its impact is relatively high. This would require modifications to all content, including localized ones. Additionally, all localization development branches would need to be synchronized and any merge conflicts resolved.

As for the current "what it is" header, there likely was a good reason when it was introduced. I can't be certain, but I presume it was intended to define a term using just three sections. On another note, the text under "what it is" seems to provide an insufficient overview of a term (document), likely because it intentionally omits content from other headers (problem, helps).

Given the potential implications, I'd like to bring this issue (the PR) to the glossary-maintainers channel for additional input.

nate-double-u:
I agree, bringing it to the maintainers is a good idea, and you're quite correct, this'll be a big change that hits a lot of pages.

> As for the current "what it is" header, there likely was a good reason when it was introduced. I can't be certain, but I presume it was intended to define a term using just three sections. On another note, the text under "what it is" seems to provide an insufficient overview of a term (document), likely because it intentionally omits content from other headers (problem, helps).

My main issue with it is that it's awkwardly phrased as is, so I think that it should be changed or removed.

Changing it may actually be easier on the localization teams than removing it, provided the translations doesn't share that awkwardness (and hopefully they don't). Removing it guarantees that we hit almost every page on the site, but changing it may only affect the English content.

rolfedh (approved):
I support this change. I believe it improves the content.
I also support removing the first subheading for minimalism.
See https://github.com/cncf/glossary/issues/2242#issuecomment-1688844358

iamNoah1:
As the lazy consensus it to remove the what it is header completely, closing this PR

PR #2198, opened 2023-07-06
Remove links to 'Deprecated' / 'Feedback Appreciated' terms

jihoon-seo:
### Describe your changes

From https://github.com/cncf/glossary/issues/2031#issuecomment-1622093398:
> the containerization text refers to the text of "container-image" [https://github.com/cncf/glossary/blob/main/content/en/containerization.md](https://github.com/cncf/glossary/blob/main/content/en/containerization.md?rgh-link-date=2023-07-05T16%3A22%3A27Z)
Does it mean that we have to remove the link to that page in our translation ?

Currently these terms are in 'Feedback Appreciated' status:
- container-image
- digital-certificate
- gitops
- hypervisor
- load-balancer
- stateless-apps

And these are in 'Deprecated' status:
- containers-as-a-service
- database-as-a-service
- debugging
- firewall
- managed-services
- platform-as-a-service
- software-as-a-service
- version-control

This PR removes all the links to the 'Deprecated' or 'Feedback Appreciated' terms, and to their localized versions.

### Related issue number or link (ex: `resolves #issue-number`)

https://github.com/cncf/glossary/issues/2031#issuecomment-1622093398

### Checklist before opening this PR (put `x` in the checkboxes)
- [X] This PR does not contain plagiarism
  - don’t copy other people’s work unless you are quoting and contributing it to them.
- [X] I have signed off on all commits 
  - [signing off](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) (ex: `git commit -s`) is to affirm that commits comply [DCO](https://wiki.linuxfoundation.org/dco). If you are working locally, you could add an alias to your `gitconfig` by running `git config --global alias.ci "commit -s"`.

PR #2353, opened 2023-09-05
Update Spanish reference links

### LINKED ISSUE #2266: [es] Update references between concepts
We want to have the concept references updated between the documents to refer to others correctly.

jossemarGT:
### Describe your changes

Since not all the glossary terms were available at the same time, contributors deferred from referencing terms in Spanish. Now, that we have the majority of terms translated we need to update those to the correct ones.

For the record this can be easily spotted with the following grep expressions:

```sh
# Needs GNU grep
grep -PHnro "\[.*?\].*?\(/.*?\)" content/es | grep -v '(/es/'
```

I double checked that each referenced `/es` term exist, however, there are 2 "special" cases I need to bring to your attention:

1. I had omitted the SaaS term since it's deprecated and its references were removed in main branch already [PR](https://github.com/cncf/glossary/pull/2198) & [commit](https://github.com/cncf/glossary/commit/1d461662ae35c8d1d3b13eb87f42e8be2835143c#diff-e39259e7b9696cdc87944e88a1a570ab38112e4d782c4e3f2502fffeaef843fd)
2. I updated `tightly-coupled-architectures` references as well, although the concept only exist on `dev-es` branch at the time I created this PR. 

Resolves #2266 

### DIFF
--- a/content/es/auto-scaling.md
+++ b/content/es/auto-scaling.md
@@ -5,7 +5,7 @@ category: Propiedad
 tags: ["infrastructure", "", ""]
 ---
 
-El autoescalado es la habilidad de un sistema para [escalar](/scalability) automáticamente, en términos de recursos computacionales.
+El autoescalado es la habilidad de un sistema para [escalar](/es/scalability) automáticamente, en términos de recursos computacionales.
 Con un sistema de autoescalado, los recursos son agregados automáticamente cuando se necesitan y pueden escalar para cumplir con la demanda fluctuante de los usuarios.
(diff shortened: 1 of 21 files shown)


electrocucaracha (approved):
/lgtm

ramrodo (approved):
Great! Thanks, @jossemarGT!

About your two points:
1. Agree to omit the SaaS term
2. I see both Concepts on both branches: [main](https://github.com/cncf/glossary/blob/main/content/en/tightly-coupled-architectures.md) & [dev-es](https://github.com/cncf/glossary/blob/dev-es/content/es/tightly-coupled-architectures.md). Not sure if your local fork/branch is updated from upstream.

PR #2366, opened 2023-09-10
[it] Update it.toml

ar406:
Some parts were still in English. I translated them into Italian.

### Describe your changes

I translated into Italian some portions that were still in English.

### DIFF
--- a/i18n/it.toml
+++ b/i18n/it.toml
@@ -20,21 +20,21 @@ other = "in"
 
 # Phrases for tags
 [ui_see_all_tags]
-other = "See all tags"
+other = "Visualizza tutti i tag"
 [ui_tag]
 other = "Tag"
 [ui_tags]
-other = "Tags"
+other = "Tag"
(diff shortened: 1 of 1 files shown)


ugho16:
Hi @ar406, 
first of all, thank you for your contribution! Unfortunately, I think we cannot accept this PR because the process we usually follow is different. PRs related to a particular language should be opened to the correct localization branch, and not to the `main` branch. In this case, you should use the branch `dev-it`. 
However, the `dev-it` branch is currently a little behind `main`, so you will not find there the changes you want to apply with this PR. We are planning to open a PR to push all the current edits in `dev-it` to `main` these days, so you will find them shortly.

In the meantime, please consider joining our [Slack channel](https://cloud-native.slack.com/archives/C02M9HPU9T9) to meet the Italian localization team. We are always happy to welcome new members!

sistella:
close as per latest comment

PR #2411, opened 2023-09-17
Merge dev-de in main

iamNoah1:
### Describe your changes

In order to get the completed terms online, we want to merge dev-de into main

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -68,6 +68,7 @@ description = "The CNCF Cloud Native Glossary Project is intended to be used as 
 #time_format_default = "02.01.2006"
 #time_format_blog = "02.01.2006"
 
+
 [languages.hi]
 title = "क्लाउड नेटिव शब्दावली"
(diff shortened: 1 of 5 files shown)


seokho-son (commented):
Hi @iamNoah1 🙂

I believe we need to use this link https://github.com/cncf/glossary/compare/main...dev-de to create a PR for updating main from a dev branch. (especially since there are no conflicts)

By utilizing upstream branches, we can minimize merge conflicts among different branches and streamline the content review process. (I think there is no need to include `config.toml` file here.)

jihoon-seo:
I agree with @seokho-son. We can directly merge `dev-de` into `main`, since there is no merge conflicts at this moment.
BTW, when merging `dev-de` into `main`, it would be great to use `Squash and merge` method, since the number of new German terms is 4, but the number of commits is 77, which seems quite redundant.

iamNoah1:
sure thing, I will go ahead and close this one.

PR #2439, opened 2023-09-30
Update virtualization.md

kumarankit999:
Updated the Problem Address by virtualization

### Describe your changes
I have corrected the word whilst by while, and also added some problems which are been addressed by virtualization.

- [ x] This PR does not contain plagiarism
- [ x] I have signed off on all commits 

### DIFF
--- a/content/en/virtualization.md
+++ b/content/en/virtualization.md
@@ -21,7 +21,7 @@ As an example, you can lease a "computer" from AWS – that computer is actually
 
 Virtualization addresses a number of problems, including the improvement of physical hardware usage 
 by allowing more apps to run on the same physical machine 
-whilst still being isolated from each other for security.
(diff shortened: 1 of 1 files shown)


content/en/virtualization.md
```diff
@@ -21,7 +21,7 @@ As an example, you can lease a "computer" from AWS – that computer is actually
 
 Virtualization addresses a number of problems, including the improvement of physical hardware usage 
 by allowing more apps to run on the same physical machine 
-whilst still being isolated from each other for security.
+while still being isolated from each other for security. Also, with virtualization, all the servers and softwares required by other cloud providers are managed by third parties, and cloud providers have to pay money monthly or annually.
```
jihoon-seo, line 24: Thank you for your contribution, but I don't get it.
The suggested content might be correct for some cases, but for other cases it would be not.
Also I think that in here we don't need to mention that cloud providers have to pay to someone else.

iamNoah1: yeah, I have to agree with Jihoon. I don't think it really adds any value :/ sorry to say that @kumarankit999

PR #2701, opened 2023-12-11
Sorting existing languages in the menu

### LINKED ISSUE #2480: Localizations order in the menu
Now that we have published numerous Glossary localisations, I think it would be better to sort them alphabetically.

Currently, when users click on "English" to display the top-right menu, they get the _Français (French)_ option last since it was the latest translation added. However, I guess it's not what most users expect when looking for their specific language in the list. The bigger this list becomes, the more difficult it will be for everyone to locate the needed language.

As far as I understand, the `weight` parameter in [config.toml](https://github.com/cncf/glossary/blob/main/config.toml) is used to set the order. I'm happy to make relevant changes if the idea itself sounds reasonable to others.

shurup:
### Describe your changes
As discussed before (see the issue below), ordering the localisations in the top-right menu seems a good idea. However, while implementing it, I realised that sorting localisations written in various languages can be quite tricky.

Based on [this ICU demo](https://icu4c-demos.unicode.org/icu-bin/locexp?_=en_US&d_=en&x=col) and some common sense (e.g. looking at how Google sorts the languages in its interfaces), I ended up with this order for existing languages:

- English (should be always first anyway)
- Deutsch
- Español (Spanish)
- Français (French)
- Italiano
- Português
- हिन्दी (Hindi)
- বাংলা (Bengali)
- 한국어 (Korean)  
- 简体中文 (Simplified Chinese)
- 繁體中文 (Traditional Chinese)
- اردو (Urdu)

I'll be happy to see any other thoughts on the best order (if there are any).

One last important thing to be aware of: we will have to update various `weight` values for various languages with each new localisation added (since this new language might have a higher position in the resulting menu).

Resolves #2480

### DIFF
--- a/LOCALIZATION.md
+++ b/LOCALIZATION.md
@@ -57,25 +57,30 @@ Open a PR with the localization initiation following this example: https://githu
 
 To add a new language to the site, modify [config.toml](https://github.com/cncf/glossary/blob/main/config.toml#L54) (Note: localization teams should use their assigned development branch for this).
 
-The `[languages]` block of `config.toml` is used to set the language. For instance, `[languages.en]` stands for English and `[languages.ko]` for Korean language configuration. Go to the `[languages]` block in `config.toml` and add a new block for your language-specific configuration. For instance, the Korean localization team added its `[languages.ko]` block after the `[languages.en]` block.
+The `[languages]` block of `config.toml` is used to set the language. For instance, `[languages.en]` stands for English and `[languages.ko]` for Korean language configuration. Go to the `[languages]` block in `config.toml` and find a place for your new localization:
+
+ - The English language always stays first.
+ - Other languages are sorted based on the alphabetized English translation of their names (see how currently existing localizations are ordered as an example).
+ - Each localization is written in its own language followed by the English translation in parentheses — e.g., `한국어 (Korean)`.
+ - The `weight` parameter in the config is used for sorting (`1` is for English, `2` is for the alphabetically first language, `3` is for the second one, and so on).
+
+For instance, you're adding a new block for the Korean localization. You calculated its weight as 8. Find the language block with `weight = 7` and add your new block afterwards:
 
 - Example of `New language block for /config.toml`
   ```diff
   [languages]
-  [languages.en]
-  title = "Cloud Native Glossary"
-  description = "The CNCF Cloud Native Glossary Project is intended to be used as a reference for common terms used when talking about cloud native applications."
-  languageName ="English"
-  # Weight used for sorting.
-  weight = 1
+  ...
+  weight = 7
   
   +[languages.ko]
   +title = "클라우드 네이티브(Cloud Native) 용어집"
   +description = "CNCF 클라우드 네이티브 용어집 프로젝트는 클라우드 네이티브 애플리케이션에 대한 대화를 나눌 때 공통의 용어를 참조하여 사용하도록 하는 목적을 가지고 있다."
-  +languageName ="한국어(Korean)"
+  +languageName = "한국어 (Korean)"
   +contentDir = "content/ko"
-  +weight = 2
+  +weight = 8
   ```
+
+Since the `weight` values should always be kept consistent, please **remember to increment the weights** of all the languages following your newly added localization.
 
 #### 3-2. Adding a file for site strings
 
--- a/config.toml
+++ b/config.toml
@@ -68,85 +68,85 @@ description = "The CNCF Cloud Native Glossary Project is intended to be used as 
 #time_format_default = "02.01.2006"
 #time_format_blog = "02.01.2006"
 
-[languages.hi]
-contentDir = "content/hi"
-languageName = "हिन्दी (Hindi)"
-title = "क्लाउड नेटिव शब्दावली"
-weight = 2
-[languages.hi.params]
-description = "CNCF क्लाउड नेटिव शब्दावली परियोजना का उद्देश्य क्लाउड नेटिव एप्लिकेशन के बारे में बात करते समय प्रयोग किए जाने वाले सामान्य शब्दों के संदर्भ के रूप में उपयोग किया जाना है।"
-
-[languages.pt-br]
-contentDir = "content/pt-br"
-languageName = "Português"
-title = "Glossário Cloud Native"
-weight = 3
-[languages.pt-br.params]
-description = "O Projeto CNCF Glossário Cloud Native se destina a ser usado como uma referência para termos comuns usados ao falar sobre aplicações nativas de nuvem."
-
-[languages.it]
-contentDir = "content/it"
-languageName = "Italiano"
-title = "Glossario Cloud Native"
-weight = 4
-[languages.it.params]
-description = "Il progetto del Glossario CNCF Cloud Native è pensato per essere un riferimento per i termini comuni, usati nell'ambito delle applicazioni cloud native."
-
-[languages.ko]
-contentDir = "content/ko"
-languageName = "한국어 (Korean)"
-title = "클라우드 네이티브(Cloud Native) 용어집"
-weight = 5
-[languages.ko.params]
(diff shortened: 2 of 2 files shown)


seokho-son (commented):
Hi @shurup 
Thanks for implementing https://github.com/cncf/glossary/issues/2480 :)

I agree with the current sorting method. (actually, I think the order depends on personal preferences)

Since you've sorted based on English words, it would be better to add (complete) English word for each language title to improve consistency.

languageName = "Português (Portuguese)"
languageName = "Italiano (Italian)"
languageName ="Deutsch (German)"

If German is acceptable, I guess we need to change the order for Deutsch

shurup:
> it would be better to add (complete) English word for each language title to improve consistency.

Great idea indeed, thank you!

> Since you've sorted based on English words

Just to clarify this moment: currently, the sorting is based not on the English words, but on some kind of transliteration of original language names into English. It means we consider _Español (Spanish)_ as a language name starting from _E_ (not from _S_). That's why _Deutsch_, which starts from _D_, seems to be in place — i.e., despite adding its English name after the original form, we still consider _Deutsch_ as the main word for sorting.

seokho-son:
Hi @shurup, my assumption about sorting by English was a misunderstanding. ;)

However, still, I believe having a clear and consistent criterion for ordering is preferable. From this perspective, I'd like to suggest listing based on the English words written in parentheses. This is important as we also need to update the guidelines accordingly.

In fact, as you mentioned, languages like Español and Deutsch can be read in English, so matching the sorting criteria to this makes sense. However, considering other languages and potential new additions, I want to avoid any special complications.

After all, local users will quickly find their language regardless of the order (for example, Koreans will easily spot `한국어`), and they are unlikely to search based on the English spelling in the parentheses. If the sorting is done in English, primarily for the benefit of English-speaking locals (mainly in the US), I don't see a major issue.

What do you think?

I'd like hear @nate-double-u and @iamNoah1 's opinion as well. :)

nate-double-u:
Do we need the English in parentheses? We can order it that way if we'd like, but i'm not sure we need the language name in English. I'm thinking about who the menu item is for -- English speakers will see "English" in English, so will be able to navigate if needed, and other language readers will see their language written in their language, so I'm not sure what the value of having each language's name listed in English would be.

I'm happy for the sorting to be done with this transliteration process, but we should update the https://github.com/cncf/glossary/blob/main/LOCALIZATION.md file with the new process.

iamNoah1:
I agree with sorting based on english term, but also not sure if we need to add it in parentheses. But I have also no strong opinion about the later.

Okabe-Junya:
Hi there!!

FYI: In kubernetes/website repo as well, a similar PR has been created.

- https://github.com/kubernetes/website/pull/44869

If this is helpful for English speakers, I agree with merging this PR. (as son-san mentioned, many Japanese would look for "日本語" too).

Additionally, if we make this change, it seems more understandable to have config.toml in the order of weights. In other words, it seems easier to read if it is as follows:

```toml
[languages.foo]
...
 weight = 2

[languages.bar]
...
weight = 3
```

shurup:
Thank you everyone for your thoughts!

Personally, I think that following the English words to order languages **and** removing English in parentheses doesn't make much sense. Based on those criteria, we'll get such a list:

- বাংলা
- English _(or it might still go the first one)_
- Français
- Deutsch
- हिन्दी
- Italiano
- 한국어
- Português
- 简体中文
- Español
- 繁體中文
- اردو

… which seems to be quite random to me as a person who can't read any non-Latin symbols. (And I guess it will still be pretty awkward for those who can read more languages, but definitely not all of them.) Adding the English names in parentheses helps simply because I start seeing some logic behind this ordering :rofl: 

At the same time, perhaps the _transliteration_ of all languages is too difficult to explicitly explain and maintain (e.g., writing guidelines, etc.), which I can fully understand.

Thus, I think that sorting the English names and specifying those English names in parentheses is the best compromise. As I can see from the Kubernetes documentation (thanks, @Okabe-Junya, for sharing!), the same logic was applied there.

I also like the idea of reordering the language blocks in `config.toml` according to the newly determined UI-facing weights of languages.

Does this all sound reasonable to others? If so, I'm ready to make corresponding changes, including those in the "Initiating a new localization team" docs. Otherwise, let's decide which way will be better.

seokho-son:
Hi @nate-double-u @iamNoah1 @Okabe-Junya @shurup 

Thank you to everyone who participated in the discussion. (And a special thanks to @shurup for summing it up so well)
It seems we've reached a general consensus through our discussion. As I mentioned initially, given that the Glossary is viewed by audiences speaking various languages, it seemed unlikely that there would be a clear-cut sorting method. In open source, even if there's no single right answer, the best solution often emerges through discussion. :) Thanks again.

Here's a summary, and it would be great if @shurup could confirm and implement these:

- Add parentheses after the language names to include the Latin transcription (to be added for languages currently not having this).
- Sort each language based on the alphabetized transcription using the 'weight' in the config (but let's keep English first, as it's the source of the Glossary definitions).
- Arrange the order of language settings in the config also according to their weight.


Additional points for implementation:

- While resolving merge conflicts in PRs, add Japanese as a new language.
- Update the method for adding new language settings to the site configuration on https://github.com/cncf/glossary/blob/main/LOCALIZATION.md#3-1-adding-a-new-language-setting-to-site-configuration (currently, languages are added at the end).

shurup:
Thanks for this excellent summary (and your very detailed ToDo list), @seokho-son! I'll make relevant changes and let you know here.

shurup:
Hi @seokho-son! All changes are done. Please review them and let me know if something needs to be corrected.

Conflicts are also resolved now.

LOCALIZATION.md
```diff
@@ -57,26 +57,31 @@ Open a PR with the localization initiation following this example: https://githu
 
 To add a new language to the site, modify [config.toml](https://github.com/cncf/glossary/blob/main/config.toml#L54) (Note: localization teams should use their assigned development branch for this).
 
-The `[languages]` block of `config.toml` is used to set the language. For instance, `[languages.en]` stands for English and `[languages.ko]` for Korean language configuration. Go to the `[languages]` block in `config.toml` and add a new block for your language-specific configuration. For instance, the Korean localization team added its `[languages.ko]` block after the `[languages.en]` block.
+The `[languages]` block of `config.toml` is used to set the language. For instance, `[languages.en]` stands for English and `[languages.ko]` for Korean language configuration. Go to the `[languages]` block in `config.toml` and find a place for your new localization:
+
+ - the English language always stays first;
```
iamNoah1, line 62:
```suggestion
 - the English language always stays first.
```

LOCALIZATION.md
```diff
@@ -57,26 +57,31 @@ Open a PR with the localization initiation following this example: https://githu
...
+ - other languages are sorted based on the alphabetized English translation of their names (see how currently existing localizations are ordered as an example);
```
iamNoah1, line 63:
```suggestion
 - other languages are sorted based on the alphabetized English translation of their names (see how currently existing localizations are ordered as an example).
```

LOCALIZATION.md
```diff
@@ -57,26 +57,31 @@ Open a PR with the localization initiation following this example: https://githu
...
+ - each localization is written in its own language followed by the English translation in parentheses — e.g., `한국어 (Korean)`;
```
iamNoah1, line 64:
```suggestion
 - each localization is written in its own language followed by the English translation in parentheses — e.g., `한국어 (Korean)`.
```

iamNoah1 (approved):
thanks @shurup LGTM

shurup:
Thanks, @iamNoah1! I applied your changes: dots instead of semicolons at the end of lists. My sense of beauty also suggested I should then capitalise the first letters in this list, which I did.

I also rebased this PR to `main` and fixed the order for the newly added Turkish localisation. If everything is good, it would be fantastic to merge it and avoid further rebasing/reordering for new languages :sweat_smile: :pray:

seokho-son (approved):
+1 approval to help @iamNoah1 make final decision. :)

PR #2801, opened 2024-01-15
Installs GTM instead of just GA4

cjyabraham:
To allow us more flexibility we want to install GTM in place of just GA4. We will also be including a HubSpot tracking code using GTM. See https://github.com/cncf/cncf.io/issues/823

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -4,6 +4,10 @@
     {{ partial "head.html" . }}
   </head>
   <body class="td-{{ .Kind }}">
+    <!-- Google Tag Manager (noscript) -->
+    <noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-WJJ7VKZ"
+    height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
+    <!-- End Google Tag Manager (noscript) -->
     <header>
       {{ partial "navbar.html" . }}
     </header>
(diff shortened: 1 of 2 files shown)


layouts/_default/baseof.html
```diff
@@ -4,6 +4,12 @@
     {{ partial "head.html" . }}
   </head>
   <body class="td-{{ .Kind }}">
+    {{ if hugo.IsProduction }}
```
jihoon-seo, line 7: From this line, I expected that the lines L9-L10 would not appear in Netlify preview (because it's not prod),
but in the source of the [preview](https://deploy-preview-2801--cncfglossary.netlify.app), those lines are rendered.

My questions are:
1. Is this right? Does Hugo presume Netlify previews as prod?
   (Or, it's just Hugo cannot tell Netlify previews and the real prod (https://glossary.cncf.io) apart?)
2. Would be there no problem if GTM tags are inserted in Netlify preview sites?

cjyabraham: I think it's not correct. Previews should not have analytics and should not be `true` for `IsProduction`.  It's not a huge concern, however, as it has been how this site has been working from the beginning with Google Analytics and GTM should not be any different.

I suggest we split this problem out as an independent issue and not allow it to hold up merging this PR?

seokho-son (approved):
LGTM :)
Thanks @cjyabraham !

layouts/_default/baseof.html
cjyabraham: Here's [a new issue for this problem](https://github.com/cncf/cncf.io/issues/824).

jihoon-seo (approved):
LGTM! Thank you!

PR #2810, opened 2024-01-16
Try to explicitly set HUGO_ENV in netlify.toml

cjyabraham:
To address [this issue](https://github.com/cncf/cncf.io/issues/824), we want to explicitly set `HUGO_ENV` so that it only is "production" on the production instance of the site. This is important so we don't run Google Analytics on preview or dev instances.


jihoon-seo (approved):
I confirmed that GTM-related codes are not rendered in the source of [preview](https://deploy-preview-2810--cncfglossary.netlify.app).
LGTM!

PR #2813, opened 2024-01-17
add info about not completed terms

iamNoah1:
### Describe your changes
When opening an issue, there was only a link to the glossary to check if suggested terms already exist. But we also have terms that are not finished, thus not show up in the glossary. Contributors who open an issue for a new term should also check there and create an issue to update the term instead of an issue to add the term. 

none

### DIFF
--- a/.github/ISSUE_TEMPLATE/new-term.yml
+++ b/.github/ISSUE_TEMPLATE/new-term.yml
@@ -34,7 +34,7 @@ body:
           required: true
         - label: The term has not already been declined in the past. [(shift+click to check)](https://github.com/cncf/glossary/issues?q=is%3Aissue+is%3Aclosed+label%3A%22triage%2Fnot+accepted%22)
           required: false
-        - label: The term does not already exist in the [glossary](https://glossary.cncf.io).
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/new-term.yml
```diff
@@ -34,7 +34,7 @@ body:
           required: true
         - label: The term has not already been declined in the past. [(shift+click to check)](https://github.com/cncf/glossary/issues?q=is%3Aissue+is%3Aclosed+label%3A%22triage%2Fnot+accepted%22)
           required: false
-        - label: The term does not already exist in the [glossary](https://glossary.cncf.io).
+        - label: The term does not already exist neither in the [glossary](https://glossary.cncf.io) as completed term, nor in [this directory](https://github.com/iamNoah1/glossary/tree/main/content/en) as not completed term. In case of the latter, you should create an issue to update the term.
```
seokho-son, line 37: How about 
```suggestion
        - label: The term does not already exist in the [glossary](https://glossary.cncf.io) as a completed term, nor does it exist in [this directory](https://github.com/cncf/glossary/tree/main/content/en) as an incomplete term. (If the status of the term is `Feedback Appreciated`, consider updating the term.)
```

for simplicity :)

seokho-son (commented):
Thanks! a minor suggestion. :)

nate-double-u (approved):
Thanks for the update!

seokho-son (approved):
Thanks!

PR #2217, opened 2023-07-11
[EN] update FaaS & Serverless

johannes-manner:
### Update FaaS and Serverless definitions based on a scientific investigation of these terms

Serverless and FaaS are often used interchangeably when people talk about event-driven stateless functions.
I am a 6th year PhD student at the University of Bamberg, worked a lot on FaaS in the last years as you can see by my [published list of papers](https://www.uni-bamberg.de/pi/team/manner-johannes/).
The ambigous use of terms motivated me to make a structured assessment of scientific literature, how they use the terms FaaS and Serverless. A preprint to my paper can be found [here](https://www.researchgate.net/publication/372214619_A_Structured_Literature_Review_Approach_to_Define_Serverless_Computing_and_Function_as_a_Service).
The reference currently included in the text (Manner 2023) is the landing page (forwarded by `doi.org`) at IEEE as soon as the paper is published - I presented it last week on [IEEE CLOUD 2023](https://conferences.computer.org/cloud/2023/). 

Hope we have a nice discussion here about these two terms :)

### Checklist before opening this PR (put `x` in the checkboxes)
- [x] This PR does not contain plagiarism
  - don’t copy other people’s work unless you are quoting and contributing it to them.
- [x] I have signed off on all commits 
  - [signing off](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) (ex: `git commit -s`) is to affirm that commits comply [DCO](https://wiki.linuxfoundation.org/dco). If you are working locally, you could add an alias to your `gitconfig` by running `git config --global alias.ci "commit -s"`.

PR #2981, opened 2024-03-10
[EN] update Function as a Service

guillaumebernard84:
### Describe your changes
Update `Function as a Service` term.

Line 32 says:
*FaaS infrastructure will auto-scale the code for heavy use*

It's not code that auto-scales, but the underlying infrastructure, the number of instances which are running the code.
I think this will be more accurate:
*FaaS will auto-scale the infrastructure for heavy use,*

The french localization team noticed this while translating the term: https://github.com/cncf/glossary/pull/2973
As this is a minor fix I didn't open an issue, I opened this PR directly.

### DIFF
--- a/content/en/function-as-a-service.md
+++ b/content/en/function-as-a-service.md
@@ -29,6 +29,6 @@ The cloud infrastructure necessary to run an app is active even when the app isn
 
 FaaS gives developers an [abstraction](/abstraction/) for running web applications in response to events without managing servers. 
 For example, uploading a file could trigger custom code that transcodes the file into various formats. 
-FaaS will auto-scale the infrastructure for heavy use, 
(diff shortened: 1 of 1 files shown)


iamNoah1 (changes requested):
Hei @guillaumebernard84 and thanks for the suggestion. I get your point that the sentence was not completely correct. But it was easier to understand. Now maybe the connection between scaling the infrastructure and the app code is lost. I would suggest something like this: "FaaS will auto-scale the underlying infrastructure that runs the application code ... "

guillaumebernard84:
@iamNoah1 I agree with you, it's more accurate like that.

nate-double-u:
Thanks for this @guillaumebernard84, but https://github.com/cncf/glossary/pull/2217 looks to have made this edit obsolete.

PR #2999, opened 2024-03-22
[ja] update redirect settings in Japanese Localization

Okabe-Junya:
### Describe your changes

- add redirect rules from `/ja/microservices/` to `/ja/microservices-architecture/` (same as English)
- replace `/ja/microservices-architecture/` to `/ja/microservices/` in `contents/ja` dir

### Background

When localizing, I often found it confusing to link to microservice pages because the destination page differs between Japanese and English.

**English**

https://github.com/cncf/glossary/blob/83d8503c36b5eed1206c7b72a0c99a78f366aa95/content/en/application-programming-interface.md?plain=1#L24

**Japanese**

https://github.com/cncf/glossary/blob/83d8503c36b5eed1206c7b72a0c99a78f366aa95/content/ja/application-programming-interface.md?plain=1#L24

 I thought it would be best to update the redirect rules to resolve this issue.

### Related issue number or link (ex: `resolves #issue-number`)

No issue related to this PR

### Checklist before opening this PR (put `x` in the checkboxes)
- [x] This PR does not contain plagiarism
  - don’t copy other people’s work unless you are quoting and contributing it to them.
- [x] I have signed off on all commits 
  - [signing off](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) (ex: `git commit -s`) is to affirm that commits comply [DCO](https://wiki.linuxfoundation.org/dco). If you are working locally, you could add an alias to your `gitconfig` by running `git config --global alias.ci "commit -s"`.

PR #3009, opened 2024-03-25
[en] Update links to 'microservices'

jihoon-seo:
### Describe your changes
Updates all links that points `/microservices/` to point `/microservices-architecture/` in all English files.

This is because IMO it is more appropriate to link to the final destination page (in this case, `/lang/microservices-architecture/`), and so we don't need to link to transient URLs (in this case, `/lang/microservices/`).
And if we leave it as-is, it might cause confusion to L10n teams and other contributors (like in #2999).

### Related PR comment
Please refer to https://github.com/cncf/glossary/pull/2999#issuecomment-2017381555


Okabe-Junya (commented):
Thank you for opening this PR!! I'm not a reviewer, but LGTM, thank you!!

iamNoah1 (approved):
thanks @jihoon-seo 

lgtm

nate-double-u (approved):
Thanks for the update @jihoon-seo!

PR #3207, opened 2024-06-23
Don't store search increments in the browser history

cjyabraham:
Currently when you refine your search term, each additional edit get stored in the browser history so that clicking Back will only take you to the previous edit. This eliminates that storage so that clicking Back will take the user to the previous page.


jihoon-seo (approved):
Preview: https://deploy-preview-3207--cncfglossary.netlify.app/search/?q=cloud

In preview, it works as intended. LGTM!

seokho-son (approved):
lgtm!!

PR #3210, opened 2024-06-28
Update some dependencies

### LINKED ISSUE #3189: [infra] Upgrade Docsy, postcss-cli, and autoprefixer
More recent releases of Docsy do not require recursive submodules, have updated dependencies, and other improvements. Consider upgrading.

Okabe-Junya:
### Describe your changes

- Update some dependencies

ref. #3189 

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -52,7 +52,7 @@
   {{ if hasPrefix .Site.GoogleAnalytics "G-"}}
     {{ template "_internal/google_analytics.html" . }}
   {{ else }}
-    {{ template "_internal/google_analytics_async.html" . }}
+    {{ template "_internal/google_analytics.html" . }}
   {{ end }}
 
   <!-- Google Tag Manager -->
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -49,11 +49,7 @@
 {{ partial "hooks/head-end.html" . }}
 <!--To comply with GDPR, cookie consent scripts places in head-end must execute before Google Analytics is enabled -->
 {{ if hugo.IsProduction }}
-  {{ if hasPrefix .Site.GoogleAnalytics "G-"}}
-    {{ template "_internal/google_analytics.html" . }}
-  {{ else }}
-    {{ template "_internal/google_analytics.html" . }}
-  {{ end }}
+  {{ template "_internal/google_analytics.html" . }}
 
   <!-- Google Tag Manager -->
   <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   },
   "devDependencies": {
     "autoprefixer": "^10.4.19",
-    "hugo-extended": "^0.127.0",
+    "hugo-extended": "^0.122.0",
     "postcss": "^8.4.38",
     "postcss-cli": "^11.0.0"
   },
--- a/package.json
+++ b/package.json
@@ -29,9 +29,10 @@
     "lunr-languages": "^1.9.0"
   },
   "devDependencies": {
-    "autoprefixer": "^9.8.6",
-    "hugo-extended": "^0.115.2",
-    "postcss-cli": "^7.1.2"
+    "autoprefixer": "^10.4.19",
+    "hugo-extended": "^0.127.0",
+    "postcss": "^8.4.38",
+    "postcss-cli": "^11.0.0"
   },
   "optionalDependencies": {
     "netlify-cli": "^17.26.1",
--- a/package.json
+++ b/package.json
@@ -25,9 +25,6 @@
     "test": "npm run check:links",
     "update:pkgs": "npx npm-check-updates -u"
   },
-  "dependencies": {
-    "lunr-languages": "^1.14.0"
-  },
   "devDependencies": {
     "autoprefixer": "^10.4.19",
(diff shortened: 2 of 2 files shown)


layouts/partials/head.html
```diff
@@ -52,7 +52,7 @@
   {{ if hasPrefix .Site.GoogleAnalytics "G-"}}
     {{ template "_internal/google_analytics.html" . }}
   {{ else }}
-    {{ template "_internal/google_analytics_async.html" . }}
+    {{ template "_internal/google_analytics.html" . }}
   {{ end }}
```
jihoon-seo, line 56: @Okabe-Junya Thank you!

In your suggestion, L53 (`{{ template "_internal/google_analytics.html" . }}`) takes effect regardless of whether `if` in L52 is true or not.

How about this?

[Option 1: comment out redundant codes]
```
  {{/* if hasPrefix .Site.GoogleAnalytics "G-" */}}
    {{ template "_internal/google_analytics.html" . }}
  {{/* else }}
    {{ template "_internal/google_analytics_async.html" . }}
  {{ end */}}
```

[Option 2: remove redundant codes]
```
  {{ template "_internal/google_analytics.html" . }}
```

(I have not tested those code snippets. We need to check whether those codes actually works without error.)

package.json
```diff
@@ -26,15 +26,16 @@
     "update:pkgs": "npx npm-check-updates -u"
   },
   "dependencies": {
-    "lunr-languages": "^1.9.0"
+    "lunr-languages": "^1.14.0"
```
cjyabraham, line 29: I think this line can be removed as we don't use lunr for search.

layouts/partials/head.html
Okabe-Junya: Thanks!

In GA4, the Async file has been deprecated, and this is now enough to configure. (https://github.com/gohugoio/hugo/issues/10899 is a useful reference)

FYI: https://gohugo.io/templates/embedded/#google-analytics

package.json
Okabe-Junya: Thanks, fixed by https://github.com/cncf/glossary/pull/3210/commits/557049345211cdf661c56b54360810b0530d08bc

layouts/partials/head.html
Okabe-Junya: Oh, I misunderstood the Option 2 you proposed. Option 2 is the most reasonable; I will fix

Okabe-Junya: fixed by https://github.com/cncf/glossary/pull/3210/commits/9a968122d34585740284824813b1ce6ddd9c6cdb

cjyabraham (changes requested):
For some reason the [Browse by Tags](https://deploy-preview-3210--cncfglossary.netlify.app/tags/?all=true) page doesn't seem to work properly. When I do "Select All" I only see one results. Could you figure out which change you made caused this?

Okabe-Junya:
Ah, great point! Thank you for bringing that up, @cjyabraham . I've been looking into this bug, and while I don’t have a complete explanation yet, it seems to occur when updating Hugo from v0.122 to [v0.123](https://github.com/gohugoio/hugo/releases/tag/v0.123.0). Since v0.123 includes some breaking changes, it might take a bit more time to fully investigate.

For this PR, I've addressed the bug by stopping the update at v0.122, and it seems to be working well now.

cjyabraham (approved):

layouts/partials/head.html
```diff
@@ -49,11 +49,7 @@
...
 {{ if hugo.IsProduction }}
-  {{ if hasPrefix .Site.GoogleAnalytics "G-"}}
-    {{ template "_internal/google_analytics.html" . }}
-  {{ else }}
-    {{ template "_internal/google_analytics_async.html" . }}
-  {{ end }}
+  {{ template "_internal/google_analytics.html" . }}
 
   <!-- Google Tag Manager -->
   <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
```
chalin, line 55: @cjyabraham - is GTM still being used by this site?

layouts/partials/head.html
chalin, line 1: Whenever Docsy is update in a followup PR, consideration should be given to eliminating this partial override, and moving any customization to, say, `layouts/partials/hooks/head-end.html` if that makes sense (or maybe a custom `layouts/partials/script.html` file).

/cc @cjyabraham

chalin (commented):
> ### Related issue number or link (ex: `resolves #issue-number`)
> close #3189

This **contributes** to #3189 **_without_ closing** it -- please update the opening comment to reflect this. Issue #3189 is also about _updating Docsy_, which is where the heavy lifting will occur. The current version of Docsy used in this repo is 3 years old.

If @cjyabraham is happy with this update, I'm willing to give it a thumbs up too.

Okabe-Junya:
Thank you for your comment, @chalin! As you suggested, I've removed the link to the Issue, so 3189 won't be closed when this PR is merged.

Additionally, I'd like to merge this PR as it is, separate from the docsy update. I think it’s useful as a maintenance PR, so I'd like to go ahead with merging it, thanks!

layouts/partials/head.html
cjyabraham: yes

Okabe-Junya:
/cc @seokho-son @jihoon-seo 

I think this PR is ready for review / merge! Please take another look to move forward, thanks!

jihoon-seo (approved):
LGTM! Thank you!

PR #3317, opened 2024-10-06
[EN] update scalability

azarboon:
### Describe your changes

I believe this term should be elaborated further. I've added more explanation. I've been inspired by Scalability definition in Reactive Manifesto: https://www.reactivemanifesto.org/glossary#Scalability. 

### DIFF
--- a/content/en/scalability.md
+++ b/content/en/scalability.md
@@ -6,7 +6,7 @@ tags: ["fundamental", "property", ""]
 ---
 
 Scalability refers to how well a system can grow. 
-That is increasing the ability to do whatever the system is supposed to do. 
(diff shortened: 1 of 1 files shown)


iamNoah1 (commented):
Thanks @azarboon for the contribution. My personal opinion is that the changes do not really add value to a simple definition of scalability in order to understand what it is in its essence. 

I am happy to hear other opinions @seokho-son @jihoon-seo @nate-double-u

nate-double-u:
Hi @azarboon, sorry for the delay in my review.

I agree with @iamNoah1 in this case. I don't think that this addition helps to simplify the definition or further explain the term.

PR #3321, opened 2024-10-12
Update container.md

azarboon:
### Describe your changes

Marc Brooker has written [a blog post](https://brooker.co.za/blog/2023/06/19/container.html) explaining that "container" means different thing to different people, and that it should be checked whether audience and narrator use the same definition. Below are his four suggested definition. Given the authority of CNCF, I suggest this nuance to be mentioned. So readers are more informed and take cautionary acts to ensure they are synced with their audience. I've added it. 

> - An approach to packaging an application along with its dependencies (sometimes a whole operating system user space), that can then run on a minimal runtime environment with a clear contract.
> - A set of development, deployment, architectural, and operational approaches built around applications packaged this way.
> - A set of operational, security, and performance isolation tools that allow multiple applications to share an operating system without interfering with each other. On Linux, this tools include chroot, cgroups, namespaces, [seccomp](https://man7.org/linux/man-pages/man2/seccomp.2.html), and others.
> - A set of implementations of these practices (the proper nouns, Docker, Kubernetes, ECS, etc).

### DIFF
--- a/content/en/container.md
+++ b/content/en/container.md
@@ -10,6 +10,8 @@ The files available to the container process are packaged as a container image.
 Containers run adjacent to each other on the same machine, 
 but typically the operating system prevents the separate container processes from interacting with each other.
 
+Note that "container" is an overloaded term. Whenever you use the word, check whether your audience uses the same definition.
+
 ## Problem it addresses
 
 Before containers were available, separate machines were necessary to run applications. 


iamNoah1 (commented):
@azarboon thanks for the contribution. Here, as well as in your other contributions I am not sure if it really adds value. 

The Glossary aims for simple straight forward definitions of terms that can understood also by non tech people. The fact the terms are overloaded and people might have different definitions is something that is true for a lot of terms. I don't know if it adds any value to point that out. 

But I am also happy to hear different opinions. @seokho-son @jihoon-seo @nate-double-u

a-mccarthy:
completely new to this project, but i'd +1 @iamNoah1 comment

>The Glossary aims for simple straight forward definitions of terms that can understood also by non tech people. The fact the terms are overloaded and people might have different definitions is something that is true for a lot of terms. I don't know if it adds any value to point that out.

I'm not sure it makes sense to add this comment to every term in the glossary. But it might make sense to add a "How to use this Glossary" section somewhere that elaborates on the glossary's founding principle to keep terms to their simplest definition and what to consider when a reader wants to use a term or learn more about a given term beyond the glossary. (please disregard this comment if this content is already available in the docs 😄)

nate-double-u:
Thanks for your effort @azarboon, but as per the previous comments, I'm closing this PR.

PR #3407, opened 2025-02-27
Update packages and required Node version

chalin:
- Updates all NPM packages except for Hugo, which will need to be addressed separately
- Updates the required Node version to match the [currently active LTS](https://nodejs.org/en/about/previous-releases). (It should also correspond to the Node version required by Docsy, but doesn't atm.)
- There are no changes to the generated site files as a result of these changes.

/cc @nate-double-u

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,18 +26,18 @@
     "update:pkgs": "npx npm-check-updates -u"
   },
   "devDependencies": {
-    "autoprefixer": "^10.4.19",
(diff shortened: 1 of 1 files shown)


nate-double-u (approved):
Thanks for this @chalin 

/lgtm

jihoon-seo (approved):
LGTM! Let me merge this.

PR #3406, opened 2025-02-27
Initial release of the Sign Language feature

seokho-son:
### Describe your changes
This PR supports https://github.com/cncf/glossary/issues/3044  
It includes six sign language videos for the corresponding six terms.

The following sign language videos are based on **the contributions and efforts of the Sign Language Team and the Deaf and Hard of Hearing WG** . These videos have already been reviewed and uploaded to the CNCF YouTube channel.

- [Autoscaling](https://youtu.be/cGONmC1smaM)  
- [Containers](https://youtu.be/meUtsFU7ndo)  
- [Reliability](https://youtu.be/pQluo2FG2eA)  
- [Serverless](https://youtu.be/rbyBgXqCN2k)  
- [Service](https://youtu.be/rE5OCe3upMo)  
- [Virtual Machine](https://youtu.be/Fq-kda8kIHg)  

**note**
- I intentionally separated YouTube embedding from the newly added `{{< sign-section >}}` shortcode. Even though we can implement video embedding section in the sign-section shortcode, I'd like to utilize the built-in `{{< youtube >}}` shortcode to minimize maintenance efforts.

**note**
- I added filter rules for spellchecker. (shortcode instruction will be ignored for spellchecking)

### Related issue number or link (ex: `resolves #issue-number`)

- Umbrella issue: #3044  
- From #3194, we reviewed the layout in the preview and agreed to use the "bottom with the H2 header" style for better visibility.  
  (welcome further style improvements!)

### Checklist before opening this PR (put `x` in the checkboxes)
- [x] This PR does not contain plagiarism
  - don’t copy other people’s work unless you are quoting and contributing it to them.
- [x] I have signed off on all commits 
  - [signing off](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) (ex: `git commit -s`) is to affirm that commits comply [DCO](https://wiki.linuxfoundation.org/dco). If you are working locally, you could add an alias to your `gitconfig` by running `git config --global alias.ci "commit -s"`.

PR #3408, opened 2025-02-27
Sign-language videos proposal

chalin:
- This is an alternative proposal to #3406 by @seokho-son, which this PR uses as a base. I figured that it would be easier to submit a clean PR than to add lots of comments to 3406. @seokho-son, feel free to copy over the content.
- **This PR is best viewed by ignoring whitespace changes.**

Relative to #3406, this PR:

- Simplifies the sign-language-section shortcode so that:
  - It is written in markdown, using `##` for the heading etc
  - It directly accepts the YouTube video ID as an argument
- Updates the heading to specify that the video is in ASL.
- Drops the "info" text, but keeps the i18n key in case some locales with to include a markdown body in the section

**Previews**:

https://deploy-preview-3406--cncfglossary.netlify.app/auto-scaling/
https://deploy-preview-3406--cncfglossary.netlify.app/container/
https://deploy-preview-3406--cncfglossary.netlify.app/reliability/
https://deploy-preview-3406--cncfglossary.netlify.app/serverless/
https://deploy-preview-3406--cncfglossary.netlify.app/service/
https://deploy-preview-3406--cncfglossary.netlify.app/virtual-machine/


spellcheck.yml
```diff
@@ -17,3 +17,10 @@ matrix:
       ignores:
       - code
       - pre
+  - pyspelling.filters.context:
+      context_visible_first: true
+      delimiters:
+      - open: '{{<'
+        close: '>}}'
+      - open: '{{%'
+        close: '%}}'
```
chalin, line 26: Btw, I have no idea if this is actually required, I just copied it over from #3406.

layouts/shortcodes/sign-language-section.html
```diff
@@ -0,0 +1,19 @@
+{{/*
+
+  Usage: {{% sign-section VIDEO_ID %}}
```
nate-double-u, line 3:
```suggestion
  Usage: {{% sign-language-section VIDEO_ID %}}
```

nate-double-u (changes requested):
If we go with this PR, I'd like the same update as I mentioned in https://github.com/cncf/glossary/pull/3406#pullrequestreview-2648983469 sorted before we merge. The container video, https://deploy-preview-3406--cncfglossary.netlify.app/container/, has some unrelated background noise that should be removed before we push this live.

seokho-son:
Thanks, @chalin !

Previously, I considered distinguishing it based on Hugo's built-in YouTube feature, but the structure you suggested is much better. Actually, I initially structured it in a similar way, but I thought that the newly defined shortcode couldn't use other prebuilt shortcodes in Hugo. It turns out that's actually possible!

I'm currently testing it further while applying the suggested shortcode to the PR #3406. Since some of the additional discussion points might overlap, I think it would be best to use this PR as a reference for the layout and conduct the overall discussion and review in the existing PR as needed.

I'll incorporate the suggestions from this PR and ask for feedback again.

layouts/shortcodes/sign-language-section.html
```diff
@@ -0,0 +1,19 @@
...
+  Usage: {{% sign-language-section VIDEO_ID %}}
+
+  */ -}}
+
+{{ $youtube := "" -}}
+{{ $videoID := .Get 0 -}}
+{{ if $videoID -}}
+  {{ $youtube = printf "{{< youtube %s >}}" $videoID }}
+{{ else -}}
+  {{ warnf "%s: Shortcode 'sign-section' requires a video ID as argument" .Page.Path -}}
```
seokho-son, line 12: Do you think we should handle the following warning due to the use of `.Page.Path` ? :)

```
WARN  .Path when the page is backed by a file is deprecated. We plan to use Path for a canonical source path and you probably want to check the source is a file. To get the current behaviour, you can use a construct similar to the one below:

  {{ $path := "" }}
  {{ with .File }}
        {{ $path = .Path }}
  {{ else }}
        {{ $path = .Path }}
  {{ end }}
```

layouts/shortcodes/sign-language-section.html
```diff
@@ -0,0 +1,19 @@
...
+{{ end -}}
+
+## {{ i18n "sign_language_header" }}
+
+{{ trim " " (i18n "sign_language_info") }}
```
seokho-son, line 17: This matter may require discussion with the sign language team (and WG) as well. Originally, there was a proposal to explicitly display the existing text, and consensus was reached on that.

At this point, I think it's best to keep it as is and consider it as a future improvement. However, I'll go through further discussions in #3406.

From a shortcode layout perspective, I’d like to ask for your opinion: If we keep the header and text as they are, would the following change be appropriate?

```diff
-{{ trim " " (i18n "sign_language_info") }}
+{{ i18n "sign_language_info" }} 
```

seokho-son (commented):
@chalin, I have the following questions regarding some technical matters. :)

spellcheck.yml
seokho-son: FYI @chalin 

This section is just to avoid spellcheck for texts inside shortcodes. (if we don't have this configuration, youtube id will be assumed as typo) 

:)

layouts/shortcodes/sign-language-section.html
chalin: Yes, fixed, thanks.

layouts/shortcodes/sign-language-section.html
chalin: Agreed, and fixed.

PR #3422, opened 2025-03-18
[editorial] Use slack.cncf.io as canonical URL

chalin:
- For context, see https://github.com/cncf/techdocs/pull/300
- Uses slack.cncf.io instead of communityinviter.com/apps/cloud-native/cncf
- Other minor copyedit.

/cc @nate-double-u


nate-double-u (approved):
LGTM, thanks for this @chalin

seokho-son (approved):
LGTM. Thanks.

Issue #3463, opened 2025-04-18
Update auto labeler workflow setting to enable for lang/vi


seokho-son:
We are initiating Vietnamese localization.

So, we need to update 

https://github.com/cncf/glossary/blob/main/.github/labeler.yml

to enable auto labeler ci for `lang/vi` for 'content/vi/**'

Good first issue. :)

PR #3460, opened 2025-04-16
Update CODEOWNERS to initialize vi l10n

Okabe-Junya:
### Describe your changes

- Update CODEOWNERS to initialize vi l10n

### Related issue number or link (ex: `resolves #issue-number`)

ref. #3458 

### Checklist before opening this PR (put `x` in the checkboxes)
- [x] This PR does not contain plagiarism
  - don’t copy other people’s work unless you are quoting and contributing it to them.
- [x] I have signed off on all commits 
  - [signing off](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) (ex: `git commit -s`) is to affirm that commits comply [DCO](https://wiki.linuxfoundation.org/dco). If you are working locally, you could add an alias to your `gitconfig` by running `git config --global alias.ci "commit -s"`.

PR #3464, opened 2025-04-18
Add auto labeler workflow setting to enable for lang/vi

ntheanh201:
### Describe your changes
Add auto labeler workflow setting to enable for lang/vi

Closes: #3463 


Okabe-Junya (commented):
Thanks, the changes are LGTM!

Question: Is it okay to use `dev-vi` as the base branch? If so, when will it be synced with main?

ntheanh201:
> Thanks, the changes are LGTM!
> 
> Question: Is it okay to use `dev-vi` as the base branch? If so, when will it be synced with main?

I see that you guys talked about how we use the localization branches, when'll we use `main` instead of `dev-{lang}`?

Okabe-Junya:
In cncf/glossary, localization teams generally open PRs against the `dev-<lang>` branch. 
This remains the case even after the initial release.

IMO: However, I find it a bit odd that GitHub Actions defines different workflows for each branch. 
It’s not a big deal for simple workflows like actions/labeler, but I’d like to sync with main and the other branches sooner rather than later.

@seokho-son 

What do you think about this? In any case, the changes are LGTM, so shall we move forward with this PR and revisit the discussion afterward?

seokho-son:
@ntheanh201 @Okabe-Junya 

Usually, the correct target branch for this kind of PR is main (we apply the most important settings to the main branch first, and then update the dev-xx branches accordingly).
However, when setting up a new localization branch, the base can be adjusted as needed.

For this case, please change the base branch of the feature branch in this PR.
This is because we need to update the dev-vi branch soon after the CODEOWNERS file is updated. https://github.com/cncf/glossary/pull/3460

Okabe-Junya:
Wow, It looks like they were submitted at exactly the same time!!

As Seokho‑son suggested, please use the main branch as the base branch :)

ntheanh201:
I changed the base branch to `main`

Okabe-Junya (approved):
Thanks, LGTM!

seokho-son (approved):
lgtm. let me merge this pr. :)

Issue #3468, opened 2025-04-21
[Update term] `Service Mesh` for `En`


jihoon-seo:
#### Why this term needs to be updated? (briefly)
- This issue is a follow-up of #3436.

We (the maintainers of the Cloud Native Glossary) decided not to add the term 'Sidecarless Service Mesh' to the Cloud Native Glossary, but instead think it would be a good enhancement to add it to the 'Service Mesh' terminology page.

Here is a guideline on how to handle this issue:
- add a brief mention of 'Sidecarless Service Mesh' (along with the opposite concept of 'Sidecarless Service Mesh', which would have a link to the term 'Sidecar Container') to the 'How it helps' section of the 'Service Mesh' glossary page

### 2. Checklist

- [x] I understand that maintainers will confirm if the term should be updated. (Please be patient)
- [ ] I want to work on this suggestion and confirm that I meet the criteria as outlined in "[who can contribute](https://glossary.cncf.io/contribute/#who-can-contribute)".. (If not checked, maintainers will assign another volunteer)

### (Note)

Maintainers will assign a label after assessment procedure. 
(from `triage/awaiting` to `triage/accepted` or `triage/not accepted`)

A contributor can be assigned to this issue after the issue get the `triage/accepted` label.
The assignee needs to check [Style Guide](https://glossary.cncf.io/style-guide/).

jihoon-seo:
@Sonichigo as an author of the issue #3436, if you are interested in taking this issue, please let us know and we will assign this issue to you. Thanks!

Sonichigo:
Hi @jihoon-seo I would be interested to pick this issue up! Can you please assign it to me?  Thanks!

jihoon-seo:
@Sonichigo Done! Thank you!

nate-double-u:
Resolved in https://github.com/cncf/glossary/pull/3511

PR #3474, opened 2025-04-24
[EN] update Service Mesh

Sonichigo:
## Describe your changes

This pull request enhances the documentation for service meshes by introducing and elaborating on the concept of sidecarless service mesh architecture. The changes highlight the differences between traditional sidecar-based models and sidecarless approaches, emphasizing their impact on resource efficiency, scalability, and performance.

### Additions to service mesh architecture:

* Added a new section titled "Implementation Models" to explain the traditional sidecar container model and its limitations, such as increased CPU/memory overhead and operational complexity as the mesh scales.
* Introduced the concept of sidecarless service mesh, which centralizes traffic control at the host level using technologies like eBPF, reducing resource consumption and network overhead.

### Benefits of sidecarless architecture:

* Detailed the advantages of using eBPF in a sidecarless service mesh, including reduced resource consumption, lower latency, and improved scalability as the number of microservices grows.

## Related issue number or link (ex: `resolves #issue-number`)

Resolves issue - #3468 

## Checklist before opening this PR (put `x` in the checkboxes)

### DIFF
--- a/content/en/cluster.md
+++ b/content/en/cluster.md
@@ -2,18 +2,17 @@
 title: Cluster
 status: Completed
 category: Concept
-tags: ["infrastructure", "fundamental", ""]
+tags: ["infrastructure", "fundamental", "architecture"]
 ---
 
-A cluster is a group of computers or applications that work together towards a common goal. 
-In the context of cloud native computing, the term is most often applied to [Kubernetes](/kubernetes/). 
-A Kubernetes cluster is a set of services (or workloads) that run in their own containers, usually on different machines. 
-The collection of all these [containerized](/containerization/) services, connected over a network, represent a cluster.
+A cluster is a group of computers or applications called [nodes](/nodes/) that work together towards a common goal.
+In cloud native computing, the term is often applied when talking about a [Kubernetes](/kubernetes/) cluster.
+It can be seen as a specific kind of [distributed system](/distributed-systems/) where the nodes are a bit more tightly coupled.
 
 ## Problem it addresses 
 
-Software that runs on a single computer presents a single point of failure 
-— if that computer crashes, or someone accidentally unplugs the power cable, 
+Software that runs on a single computer presents a single point of failure — 
+if that computer crashes, or someone accidentally unplugs the power cable, 
 then some business-critical system may be taken offline. 
 That's why modern software is generally built as [distributed applications](/distributed-apps/), grouped together as clusters. 
 
--- a/content/en/distributed-systems.md
+++ b/content/en/distributed-systems.md
@@ -2,7 +2,7 @@
 title: Distributed System
 status: Completed
 category: concept
-tags: ["architecture", "", ""]
+tags: ["architecture", "infrastructure", "fundamental"]
 ---
 
 A distributed system is a collection of autonomous computing elements 
@@ -28,4 +28,5 @@ This can be automated allowing a system to handle a sudden increase in workload 
 
 A non-distributed system exposes itself to risks of failure because if one machine fails, the entire system fails. 
 A distributed system can be designed in such a way that, 
-even if some machines go down, the overall system can still keep working to produce the same result.
+even if some machines go down, 
+the overall system can still keep working to produce the same result.
--- a/content/en/nodes.md
+++ b/content/en/nodes.md
@@ -2,14 +2,15 @@
 title: Nodes
 status: Completed
 category: Concept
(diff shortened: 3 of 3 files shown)


Sonichigo:
Hi @jihoon-seo, I have updated term and added a small section before "Problem it addresses", while following the [style-guide](https://glossary.cncf.io/style-guide/), however i couldn't understand the phrase - `to the 'How it helps' section of the 'Service Mesh' glossary page`. So do let me know, if there's any changes or structure i need to fix/correct/update.

nate-double-u (changes requested):
Thanks for this @Sonichigo. My main concern at this point is that this update will break the existing template, so edits need to be made before this change can be considered.

Sonichigo:
Make sense, thanks for the feedback. My main concern was same, as i didn't
wanted to break the template but couldn't think of much way. I'll make the
necessary changes and update asap

Sonichigo:
Hi @nate-double-u, I have merged the content from implementation  model to "How it helps" by first explaining the use of Service mesh, then it's two approaches, and adding the relevant tags to ebpf and sidecar container. Looking forward to your feedback.

nate-double-u:
A note to @Sonichigo and other reviewers, I've  taken this opportunity [to make some formatting updates](https://github.com/cncf/glossary/pull/3474/commits/1853f69c03911daf37cc0fc3c28245591383a969) to make this article conform to our request for semantic line break (one sentence per line) use to ease review. I haven't made any content updates in that commit.

Sonichigo:
The piple of spelling check is failing due to the term `Sidecarless` ilself how can i fix this? Is there a dictionary or an array where we can whitelist the term?

iamNoah1:
> The piple of spelling check is failing due to the term `Sidecarless` ilself how can i fix this? Is there a dictionary or an array where we can whitelist the term?

@Sonichigo you can add it to the wordlist.txt on the root level. But I would encourage to maybe use something: A design without sidecars.

iamNoah1:
like

content/en/service-mesh.md
```diff
@@ -5,24 +5,22 @@ category: technology
...
+Before service meshes, that functionality had to be encoded into every single service, becoming a potential source of bugs and technical debt. In the [Sidecar Container](/sidecar-container/) model, each microservice pod is paired with its own proxy.
+This per-service proxy intercepts L3–L7 traffic, enforcing load-balancing, mutual TLS, tracing and metrics on a fine-grained basis. While this approach offers maximal policy flexibility and service-specific routing, it also incurs additional CPU and memory overhead for every proxy and increases operational complexity as the mesh grows.
+
+A **Sidecarless** design, on the other hand, runs data-plane logic directly in the Linux kernel and combines mesh functionality into a small group of host-level agents that make use of kernel features like [eBPF](/ebpf/).
+By doing away with per-pod proxies, this method drastically reduces resource usage and removes unnecessary network hops, which lowers latency and boosts performance. Because overhead remains constant regardless of pod count and there are fewer agents to deploy, teams benefit from simplified operations and linear scalability as service numbers increase.
```
nate-double-u, line 26: One more semantic line break:
```suggestion
By doing away with per-pod proxies, this method drastically reduces resource usage and removes unnecessary network hops, which lowers latency and boosts performance.
Because overhead remains constant regardless of pod count and there are fewer agents to deploy, teams benefit from simplified operations and linear scalability as service numbers increase.
```

Sonichigo:
> > The piple of spelling check is failing due to the term `Sidecarless` ilself how can i fix this? Is there a dictionary or an array where we can whitelist the term?
> 
> @Sonichigo you can add it to the wordlist.txt on the root level. But I would encourage to maybe use something like: A design without sidecars.

I believe it's better to add it in wordlist.txt, since sidecarless became an joined term with release of ambient mesh last year. Wdyt?

hlipsig (approved):

nate-double-u:
Based on https://github.com/cncf/glossary/pull/3506, I'll drop the commits relating to semantic line breaks. (Will also rebase on main)

Sonichigo:
Hi @nate-double-u no worries, can you reopen the pr?

nate-double-u:
Reopen is greyed out for me. I'll open a new one.

nate-double-u:
Sorry about this folks, please see https://github.com/cncf/glossary/pull/3511.

@hlipsig @iamNoah1

PR #3524, opened 2025-05-27
[En] add Host

### LINKED ISSUE #1308: [New term] `Host`
#### What this term is about? (briefly)
- "Host" is a word that is thrown around a lot in cloud native and observability contexts, but usually lacks a concrete definition. 
#### Why this term needs to be added? (briefly) 
- "hostname" is a concept shared by containers, virtual hosts, and cloud nodes alike, which can lead to people using the word "host" to describe various types of resources and entities in cloud architectures. It would be useful for this glossary to be clear about it, one way or another.
- Much like "service" shows up in this glossary as a way to limit confusion around an over-used term (and pointing the user to "microservice" to direct that confusion in this context), it would be good to consider having a similar entry for "host". Presumably it would ultimately link to "node", but the entry would also have the opportunity to clear up the fact that "host" is almost always relative to a specific context, e.g. a pod is a host for a container, a K8S node is a host for pod, and a virtual host can be a host for a container, a microservice, or many other things. (For this reason, perhaps the term "host" should be avoided?)

I'm very curious to hear others' thoughts on this, as I'm quite new to the CNCF glossary! :)

### 2. Checklist

- [X] The suggested term fits the [cloud native definition](https://github.com/cncf/toc/blob/main/DEFINITION.md) or is essential to understanding cloud native terms. (We hope to avoid adding general or broad terminologies)
- [X] There is no existing issue yet. [(shift+click to check existing issues)](https://github.com/cncf/glossary/issues?q=label%3A%22new+term%22)
- [X] The term does not already exist in the [glossary](https://glossary.cncf.io).
- [X] I understand that maintainers will assess if the term should be part of the Glossary. (Please be patient)
- [X] I want to work on this term. (If not checked, maintainers will assign another volunteer)

### (Note)

Maintainers will assign a label after assessment procedure. 
(from `triage/awaiting` to `triage/accepted` or `triage/not accepted`)

A contributor can be assigned to this issue after the issue get the `triage/accepted` label.
The assignee needs to check [Style Guide](https://glossary.cncf.io/style-guide/).

jasonrhodes:
### Describe your changes

This has been a long time coming after I initially submitted this idea years ago!

As best as I can tell, this is similar to the "service" / "microservice" relationship, so I've tried to model this doc after the "service" one. I've added a bit of context about "host" that isn't present in "nodes", but maybe it's unnecessary? Feedback definitely appreciated.

Resolves #1308

### DIFF
--- a/content/en/host.md
+++ b/content/en/host.md
@@ -1,6 +1,6 @@
 ---
 title: Host
-status: Feedback Appreciated
+status: Completed
 category: concept
 tags: ["infrastructure", "fundamental", "architecture"]
 ---


iamNoah1 (changes requested):
I really like the simplicity of this description. However, usually we have a certain structure that we follow for this category., ie: Problem it Addresses and How it helps. But for this term I don't really see how we should press into this template ^^, I am curious if some other maintainer has an idea how to proceed. 

Appart from that the content is nice, just have a look at the failing spell and link check. 

cc @hlipsig @jihoon-seo @seokho-son @Okabe-Junya

jasonrhodes:
I've added semantic line breaks and fixed the failing "LINKS" check by adding the missing trailing slash to the nodes link. The spellcheck is failing due to the use of "hostname", which is the word I intended to use. Is there a way to specify that it's valid or does it need to be rewritten to "host name" just to pass this check?

Thanks!

nate-double-u:
> I've added semantic line breaks and fixed the failing "LINKS" check by adding the missing trailing slash to the nodes link. 

Thanks!

> The spellcheck is failing due to the use of "hostname", which is the word I intended to use. Is there a way to specify that it's valid or does it need to be rewritten to "host name" just to pass this check?

You can add it to the `wordlist.txt` file.

jasonrhodes:
> You can add it to the wordlist.txt file.

@nate-double-u got it, I've added "hostname" there, thanks.

> However, I think it's also necessary to provide a consistent style for all terminology so that users don't get confused

@Okabe-Junya I definitely agree. That was why I modeled this change after the other "redirect" style of definition in the glossary already. Please take a look at [the current service definition file](https://github.com/cncf/glossary/blob/821376ed090914b47cd415c79f0e6b88d0a21b4b/content/en/service.md). If you all would like these "redirect" definition files to include the same sections as the others, I'd like some guidance on how to approach it. Copy/paste from the target? 

Thanks.

jasonrhodes:
The last line of this new "host" file and of the mentioned "service" file both use "Please see the [target] definition", these could be expanded to say "For more about the problem it addresses and how it helps, please see the [target] definition" if that helps tie the formats together? Interested in more feedback here.

Okabe-Junya:
Thank you for pointing out, Jason, Noah!

I see... The "Problem it addresses" section is indeed very difficult. I've thought about this term many times, and I think that "host" would be a conceptual/abstract term rather than a term for solving specific problems (as you all say, that's already explained by other existing terms).

How about an option where we keep only the "How it helps" section and omit the "Problem it addresses" section?

Thank you again for your contribution and discussion :)

jihoon-seo:
IMO, while it's important to follow the established template, there may be terms that are difficult to fill in the "Problem it addresses" section and the accompanying "How it helps" section in the template, and I think the term "Host" that Jason is trying to add in this PR is one of those. So I think it would be nice to allow exceptions when necessary.. :)

Okabe-Junya:
Sorry for the late reply.

It looks like Nate and Seokho reacted "+1" for Noah’s comment. 
Let’s move this PR forward by omitting that section :)

(I'll review this PR later)

iamNoah1 (approved):
@jasonrhodes I am happy to approve now that we have a consensus amongst maintainers. Thanks again for the effort :)

PR #3631, opened 2026-04-12
fix: prevent duplicate auto-comments on PR label events

### LINKED ISSUE #3621: bug: Auto-comment workflow creates duplicate comments under certain conditions
This repo currently has a workflow enabled that adds the comment "Wait, please do not proceed with this yet." when a PR with the "hold" label is added.
- https://github.com/cncf/glossary/blob/main/.github/workflows/auto-comment.yml
- https://github.com/cncf/glossary/blob/main/.github/auto-comment-config.yaml

However, when I added the "hold" label to my PR (#3619), "github-actions" (a bot) added 15 duplicate "Wait, please do not proceed with this yet."

To find out why, I looked at https://github.com/cncf/glossary/actions/workflows/auto-comment.yml. When I added the 'hold' label, I also removed the 'lang/xx' labels that were incorrectly attached at the same time. This caused the workflow to be executed separately for each 'lang/xx' label removal event.

This behavior could be prevented by slightly modifying the workflow execution conditions in https://github.com/cncf/glossary/blob/main/.github/workflows/auto-comment.yml.

If anyone would like to take on this task, please comment on this issue to express your interest and open a pull request targeting the `main` branch.
I encourage anyone interested to contribute.

RiyaChaturvedi37:
## Problem
The auto-comment workflow was triggering on both `labeled` and `unlabeled` 
events. When multiple labels were removed at the same time as adding the 
`hold` label, the workflow ran separately for each label event causing 
duplicate comments.

## Fix
1. Removed `unlabeled` from workflow triggers
2. Added `if: github.event.label.name == 'hold'` condition to ensure 
   the workflow only runs when the `hold` label is specifically added

Fixes #3621

### DIFF
--- a/.github/workflows/auto-comment.yml
+++ b/.github/workflows/auto-comment.yml
@@ -1,54 +1,48 @@
 name: Auto Comment on Labeled PR
-
 on:
   pull_request_target:
     types:
       - labeled
-      - unlabeled
-
 permissions:
   pull-requests: write
-
 jobs:
   comment:
     runs-on: ubuntu-latest
+    if: github.event.label.name == 'hold'
     steps:
     # Checkout the repository code to the runner.
     - name: Checkout code
       uses: actions/checkout@v6
-
     # Create a bin directory in the runner's home
     - name: Create bin directory
       run: |
         mkdir -p $HOME/bin
         echo "$HOME/bin" >> $GITHUB_PATH
-
     # Install the latest jq and yq from GitHub releases
     - name: Install jq and yq
       run: |
         wget https://github.com/jqlang/jq/releases/latest/download/jq-linux-amd64 -O $HOME/bin/jq
         chmod +x $HOME/bin/jq
         wget https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64 -O $HOME/bin/yq
-        chmod +x $HOME/bin/yq      
-        
+        chmod +x $HOME/bin/yq
+
     # Main logic for adding comments based on labels.
     - name: Comment on PR based on external config
       run: |
         # Fetch the labels attached to the PR from the GitHub event JSON.
         pr_labels=$(jq -r '.pull_request.labels[] | .name' "$GITHUB_EVENT_PATH" | tr '\n' ' ')
-        
+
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-comment.yml
```diff
@@ -1,70 +1,63 @@
...
       - labeled
-      - unlabeled
-
 permissions:
   pull-requests: write
-
 jobs:
   comment:
     runs-on: ubuntu-latest
+    if: github.event.label.name == 'hold'
```
Okabe-Junya:
```suggestion
```

Since we're already filtering the trigger conditions, I don't think that's necessary.

.github/workflows/auto-comment.yml
```diff
@@ -1,70 +1,63 @@
...
+
           # If all required labels are found, add the comment to the PR.
           echo "All required labels found. Adding comment."
           gh pr comment "$pr_number" --body "$comment"
-
         done
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      shell: bash
+      shell: bash
```
Okabe-Junya: Please do not delete the newline in EOF

cf. https://pubs.opengroup.org/onlinepubs/9799919799/

> 3.172 Incomplete Line
A sequence of one or more non-<newline> characters at the end of the file.

.github/workflows/auto-comment.yml
Okabe-Junya, line 1: Please do not execute your formatter on your local, it's better to include only the minimum necessary changes in the PR

Okabe-Junya (changes requested):
Thank you for your contribution!
Left some comments, PTAL whrn you have time

RiyaChaturvedi37:
Hi @Okabe-Junya, I've addressed all the review comments — removed the redundant if condition, restored the EOF newline, and signed off the commits. Could you please re-review? Thank you!

Okabe-Junya (approved):
Thanks, LGTM

nate-double-u:
Thanks for this, @RiyaChaturvedi37, however i'm closing this PR as this issue was corrected in https://github.com/cncf/glossary/pull/3675.

PR #3561, opened 2025-09-11
Rebasing dev-tr on top of main (manually)

aliok:
### Checklist before opening this PR (put `x` in the checkboxes)
- [ ] This PR does not contain plagiarism
  - don’t copy other people’s work unless you are quoting and contributing it to them.
- [ ] I have signed off on all commits 
  - [signing off](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) (ex: `git commit -s`) is to affirm that commits comply [DCO](https://wiki.linuxfoundation.org/dco). If you are working locally, you could add an alias to your `gitconfig` by running `git config --global alias.ci "commit -s"`.

PR #3635, opened 2026-04-14
Merge main into dev-tr

aliok:
Getting updates to dev-tr, as suggested in https://github.com/cncf/glossary/pull/3561

Similar to https://github.com/cncf/glossary/pull/3575

cc @seokho-son 

PS: Is there a way to not ping all the reviewers in all the localizations for this kind of sync PRs?

Actually, since it is to be done for our (Turkish localization team's) `dev-tr` branch, we should be able to just do this ourselves. Can I directly push to that upstream/dev-tr branch w/o a PR (I haven't tried if I have the permissions)?


Okabe-Junya (approved):

Okabe-Junya:
Thank you, let me merge on behalf of maintainers :)

PR #3645, opened 2026-05-13
ci: pin contents: read on the 4 remaining PR checks

arpitjain099:
Pins the default `GITHUB_TOKEN` on the four PR-time checks still inheriting org defaults:

- `check-links.yml` — `npm run check:links`.
- `check-outdated-content.yaml` — detects out-of-date Korean/Spanish translations against upstream English content.
- `es-spellcheck.yml` — PySpelling on `content/es/`.
- `spellcheck.yml` — `rojopolis/spellcheck-github-actions` for the English content.

The sibling 4 hardened workflows in this repo already use the same top-level pattern. YAML validated locally.

PR #3649, opened 2026-05-25
ci: declare workflow-level `contents: read` on spellcheck, check-outdated-content, es-spellcheck

arpitjain099:
Three lint/check workflows just validate content. No GitHub API writes, so workflow-level `contents: read` is the right cap.

Post-CVE-2025-30066 (`tj-actions/changed-files`) hardening pattern. YAML validated locally.

### DIFF
--- a/.github/workflows/check-outdated-content.yaml
+++ b/.github/workflows/check-outdated-content.yaml
@@ -9,6 +9,9 @@ on:
     paths:
       - 'content/en/**.md'
 
+permissions:
+  contents: read
+
 jobs:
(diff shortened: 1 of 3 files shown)


Okabe-Junya (commented):
Thank you! but duplicated to https://github.com/cncf/glossary/pull/3645/changes?

arpitjain099:
Good catch @Okabe-Junya, thanks - this is indeed a subset of my earlier #3645. Closing this one out in favor of that one.

